Issue #73, opened 2022-02-14
Agent crash loop when using ADMX


SteadEXE:
### Bug reporting acknowledgment

Yes, I read it

### Professional support

Yes, I know

### Describe the bug

When deploying configuration for agent by GPO with the given ADMX, something goes wrong.

Here is a screen of registry with keys created by .msi:

![image](https://user-images.githubusercontent.com/1759156/153852132-aa23e76e-98f7-497f-a78b-68b2aa1a0360.png)

Here is a screen of registry with keys created by ADMX:

![image](https://user-images.githubusercontent.com/1759156/153852287-7c2c045f-2f7d-48a5-ae36-e5026386a8f8.png)

It looks like some keys are not getting parsed properly because when we try to repair the installation using the installer this is what get parsed:

![image](https://user-images.githubusercontent.com/1759156/153852395-ebf5e3c7-c41a-4933-84a6-f444faef1a12.png)

And the agent crash loop while trying to contact the server with a misleading error:

[Mon Feb 14 11:40:28 2022][info] GLPI Agent service starting
[Mon Feb 14 11:40:29 2022][info] GLPI Agent memory usage: WSS=3670016 PFU=82358272
[Mon Feb 14 11:40:29 2022][info] target server0: next run: Mon Feb 14 11:40:20 2022 - https://[REDACTED]/plugins/glpiinventory/index.php
[Mon Feb 14 11:40:29 2022][info] target server0: server https://[REDACTED]/plugins/glpiinventory/index.php
[Mon Feb 14 11:40:30 2022][info] sending prolog request to server0
[Mon Feb 14 11:40:30 2022][error] [http client] communication error: 500 Can't connect to [REDACTED]:443 (Une tentative de connexion a échoué car le parti connecté n’a pas répondu convenablement au-delà d’une certaine durée ou une connexion établie a échoué car l’hôte de connexion n’a pas répondu.)
[Mon Feb 14 11:40:30 2022][error] No answer from server at https://[REDACTED]/plugins/glpiinventory/index.php
[Mon Feb 14 11:40:30 2022][info] target server0: next run: Mon Feb 14 11:40:30 2022 - https://[REDACTED]/plugins/glpiinventory/index.php
[Mon Feb 14 11:40:31 2022][info] target server0: server https://[REDACTED]/plugins/glpiinventory/index.php
[Mon Feb 14 11:40:31 2022][info] sending prolog request to server0
[...]

PR #79, opened 2022-02-18
Fix parsing DWORD keys from registry on Windows

SteadEXE:
If the key is not in REG_SZ but REG_DWORD, value like "0x0000" are not read as a 0


g-bougard:
Hi @SteadEXE 

thank you for this PR. I think it is correct as you seem to have REG_DWORD in #73 where the agent expects REG_SZ for all its configuration variable.

As correct, I will probably merge it, but I would prefer also to fix the ADMX and check what could happen for the installer during an upgrade.

PR #88, opened 2022-03-23
Correct Ssh.pm for private key usage

Shugotekfr:
If no password is entered during add of remote host, glpi-agent will try to use private key available in .ssh folder of current user (in services mode, it will use root user).
But currently that not work due to error on pattern line : 
 - First, need to add "qr" to identify a Regex (See perl Regexp Quote-Like Operators )
 - Second, to search only first line and at end of first line, we must remove "$" and add "\n" at end of pattern


g-bougard:
Thank you for this good catch, I really missed it.

PR #107, opened 2022-04-28
+DefencePro Serial MIB

rajabiy:
found in
https://support.radware.com/ci/okcsFattach/get/1010467_3

### DIFF
--- a/lib/GLPI/Agent/SNMP/Device.pm
+++ b/lib/GLPI/Agent/SNMP/Device.pm
@@ -278,6 +278,7 @@ sub setSerial {
             '.1.3.6.1.4.1.232.2.2.2.1.0',            # CPQSINFO-MIB
             '.1.3.6.1.4.1.232.11.2.10.3.0',          # CPQHOST-MIB
             '.1.3.6.1.4.1.14823.2.3.3.1.2.1.1.4.144.76.129.206.143.6', # ARUBA-MIB
+            '.1.3.6.1.4.1.89.2.12.0',                # DefencePro-MIB
         );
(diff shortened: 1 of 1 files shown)


g-bougard:
Hi @rajabiy 

Anyway I would prefer another way to catch the DefencePro Serial OID. This code is really historic. Even if your code works, it means the OID will be tried on any device the OID is not found. If we extend this list more and more, this means the agent will pass more and more time to ask for wrong OID. I don't like this comportment.

So I implemented a MibSupport feature to avoid such problem: it permits to detect if we are in a context we should try this OID. Can you try to create such a module ? For example, [lib/GLPI/Agent/SNMP/MibSupport/Oki.pm](https://github.com/glpi-project/glpi-agent/blob/develop/lib/GLPI/Agent/SNMP/MibSupport/Oki.pm) was exactly developed to provide serial number for Oki printers. You can even take this one a example. You can check all other modules into MibSupport folder. Also [lib/GLPI/Agent/SNMP/MibSupportTemplate.pm](https://github.com/glpi-project/glpi-agent/blob/develop/lib/GLPI/Agent/SNMP/MibSupportTemplate.pm) can be used a developer documentation.

If you don't feel easy enough, can you just provide a [snmp walk output](https://glpi-agent.readthedocs.io/en/latest/bug-reporting.html#snmpwalk-output) and I'll propose the module for you.

Looking at the referenced document, I see the mac address can be found. Also the software version and the attack DB version could probably be added as firmware version.

rajabiy:
Thank you I will try

PR #110, opened 2022-05-03
Create DefencePro.pm

sectoolsacc:
add defencepro mib support

### DIFF
--- a/lib/GLPI/Agent/SNMP/MibSupport/DefencePro.pm
+++ b/lib/GLPI/Agent/SNMP/MibSupport/DefencePro.pm
@@ -8,10 +8,10 @@ use parent 'GLPI::Agent::SNMP::MibSupportTemplate';
 use GLPI::Agent::Tools;
 use GLPI::Agent::Tools::SNMP;
 
-use constant defencepro    => '.1.3.6.1.4.1.89';
-use constant model  => defencepro . '.2.14.0';
-use constant rndSerialNumber => defencepro . '.2.12.0';
-use constant rsWSDUserVersion => defencepro . '.35.1.34';
+use constant defencepro             => '.1.3.6.1.4.1.89';
+use constant model                  => defencepro . '.2.14.0';
+use constant rndSerialNumber        => defencepro . '.2.12.0';
+use constant rsWSDUserVersion       => defencepro . '.35.1.34';
 use constant rsWSDSysBaseMACAddress => defencepro . '.35.1.69.5.0';
 
 our $mibSupport = [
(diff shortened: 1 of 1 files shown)


lib/GLPI/Agent/SNMP/MibSupport/DefencePro.pm
```diff
@@ -0,0 +1,52 @@
...
+use parent 'GLPI::Agent::SNMP::MibSupportTemplate';
+
+use GLPI::Agent::Tools;
+use GLPI::Agent::Tools::SNMP;
+
+use constant defencepro    => '.1.3.6.1.4.1.89';
+use constant model  => defencepro . '.2.14.0';
+use constant rndSerialNumber => defencepro . '.2.12.0';
+use constant rsWSDUserVersion => defencepro . '.35.1.34';
+use constant rsWSDSysBaseMACAddress => defencepro . '.35.1.69.5.0';
```
g-bougard: Just for readability, can you align `=>` strings ?

g-bougard: Also I see you define `rsWSDUserVersion` constant but you don't use it.
Unless you still have the firmware version, you probably want to use it in a `getFirmware()` method. There's a lot of example in other MibSupport modules.
If you still have the firmware version or you want to give it a custom description, you can use `$device->addFirmware(...)` call like this is done for Digi or Epson support. You can add as many firmware as you need with custom description ;-)

lib/GLPI/Agent/SNMP/MibSupport/DefencePro.pm
```diff
@@ -0,0 +1,52 @@
...
+sub getSerial {
+    my ($self) = @_;
+
+    return $self->get(rndSerialNumber);
+}
+
+sub getMacAddress {
+    my ($self) = @_;
+
+    return $self->get(rsWSDSysBaseMACAddress);
```
g-bougard: Here it is safer to validate the MAC like in:
```suggestion
    return getCanonicalMacAddress($self->get(rsWSDSysBaseMACAddress));
```

g-bougard (commented):
Hi @sectoolsacc 
nice work.
I have very few comments and advices to enhance a little your work.
For the firmware case, if you add it, don't forget to validate this enhances the inventory.

lib/GLPI/Agent/SNMP/MibSupport/DefencePro.pm
sectoolsacc: Hello @g-bougard there are firmware version in this string something like 7.6.1 i didn't know where use it

g-bougard: Can you try to add this code somewhere and check if this looks well in glpi ?
```perl
sub getFirmware {
    my ($self) = @_;

    return getCanonicalString($self->get(rsWSDUserVersion));
}
```

sectoolsacc: ok think I did it

PR #134, opened 2022-06-03
fix: screen's edid fetch from /sys/devices

yweber-volta:
The wanted filename was matched against the absolute path : the comparison never evaluate to TRUE.
Now the filename is matched against the file's basename.

### DIFF
--- a/lib/GLPI/Agent/Task/Inventory/Generic/Screen.pm
+++ b/lib/GLPI/Agent/Task/Inventory/Generic/Screen.pm
@@ -10,6 +10,7 @@ use MIME::Base64;
 use UNIVERSAL::require;
 
 use File::Find;
+use File::Basename;
 use GLPI::Agent::Tools;
 use GLPI::Agent::Tools::Screen;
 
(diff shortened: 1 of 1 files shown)


g-bougard (commented):
OMG, you're right on that problem and I totally missed it. Thank you so much for that PR.

Indeed, the `no_chdir => 1,` line is a part of the problem as it implies `$_` will contain a full path.

Anyway I would prefer replacing `return unless $_ eq 'edid';` by a regexp check like `return unless m{^(.*/)?edid};`. What do you think ? But your code is definitively correct.

Also I think you can replace any occurrence of `$File::Find::name` by `$_` in the `wanted => sub {...}`.

yweber-volta:
Thank's for your quick reply and for the comments !
I have to admit I didn't read/write perl from a very long time ;)

After some tests, here is my proposal based on your advises : `no_chdir => 0,` allows simpler comparison etc.

g-bougard:
Indeed there's a misunderstanding: I didn't ask you to change `no_chdir`. In fact, this is a bad idea while you're running as a service.

I was okay with your initial code, but it should also replace all `$File::Find::name` occurrence then.

yweber-volta:
Sorry for misunderstanding the implication of `no_chdir` changes !

Feel free to replace the `basename` call by a regexp. Personally I'm more comfortable with relying on a well known function to process a path.

g-bougard (approved):

g-bougard:
Thank you for your nice contribution ;-)

PR #136, opened 2022-06-07
feat: Add support for linux systemd-nspawn container

### LINKED ISSUE #130: Support for Systemd-nspawn Containers in GLPI agent
Hi,
in my organization we work with Linux containers of type Systemd-nspawn. Is the GLPI agent compatible with this type of container or only with LXC? If the agent is installed on the host, the host does not detect the containers or software they contain. For example in a virtual machine there are 2 containers of type systemd-nspawn, one has an Apache web server and another a MariaDB database. The only solution we see right now is to install the agent on containers as well.
Thank you.

g-bougard:
Closes #130

I keep it open as `systemd-nspawn` is still not supported by inventory format.

### DIFF
--- a/Changes
+++ b/Changes
@@ -11,6 +11,7 @@ core:
 inventory:
 * fix Oracle inventory when ORACLE_HOME is still found in environment variables
 * fix Office License scan on win32 due to an unexpected value key
+* fix #130: Add support for linux systemd-nspawn container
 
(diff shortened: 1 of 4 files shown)


g-bougard:
Merged as GLPI 1.0.2 has been released making `systemd-nspawn` supported.

Issue #108, opened 2022-05-02
GLPI AGENT:  check ssl certificate from keystore on Windows clients.


waaalex:
### Documentation

Yes, I read it

### Is your feature request related to a problem?

issue related : 
https://github.com/glpi-project/glpi-agent/issues/105

### Expected behavior

**I would like glpi agent to validate certificate when root cert is in keystore.**
For example : We have an enterprise root authority for our domain.
glpi has his certificate, delivered by root authority.
But glpi-agent can't accept this certificate because he can't read from keystore.
We have to configure ssl certificate accept manually
(sorry for my poor english)

### Alternatives you've considered

on regedit, no-ssl-check to 1.

### Additional context

I am ok to test this feature if you can add it.
Thank you very much :)

g-bougard:
Hi @waaalex  

I prepared an update for this feature request. Do you have time to test it ?

You need to install latest nightly as the #142 PR is based on a refacto which simplify it a lot. After you have installed it, you just need to replace in your installation, the `C:\Program Files\GLPI-Agent\perl\agent\GLPI\Agent\HTTP\Client.pm` file by this one: [Client.pm](https://raw.githubusercontent.com/glpi-project/glpi-agent/a7c7a9cb0fd1218138fdbd5a4f165b1c4d620813/lib/GLPI/Agent/HTTP/Client.pm).

You can also directly try the dev MSI from the Windows-Build artifact from the [GH Actions run #2489779439](https://github.com/glpi-project/glpi-agent/actions/runs/2489779439).

In the case you're using one of `ca-cert-file` or `ca-cert-dir` options, you also need to clear it before the feature could be enabled.

To explain the purpose of the update: it will generate a file which will include re-encoded files exported with `certutil -Store -Silent -Split` command and this file will be used like it was set for `ca-cert-file` option. The file is cached for 1 hour and will be re-generated on first SSL access on glpi server after the program or service is started. So you'll also have to restart the service after the update has been applied.

Edit: As I'm using [certutil](https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2012-r2-and-2012/cc732443(v=ws.11)) command, can you also confirm you can see the expected certificate in the output of `certutil -Store -Silent` command ? Or maybe you need to add `-Enterprise` option ?

waaalex:
yes i would like to test it but i can't find link to the latest nightly. can't you give me the link?
Thanks and sorry.
[...]

PR #142, opened 2022-06-13
Add integrated keystore support for https requets

### LINKED ISSUE #33: macOS: certificate should be stored in the Keychain, not inside GLPI-Agent.app/etc
### Documentation

- [X] Yes, I do

### Professional support

Still not applicable

### Is your feature request related to a problem?

It is bad practice to require dynamic configuration to be placed within an app bundle on macOS, since this prevents signing and notarisation of the app (see also #32). In the context of certificates, macOS has a secure keystore called the Keychain. This is the correct place for certificates to be stored and read by applications. This allows the certificates to be securely deployed via MDM.

### Expected behavior

The GLPI certificate should be stored and read from the macOS Keychain.

### Alternatives you've considered

N/A.

### Additional context

_No response_

g-bougard:
Add checking server certificate against system keystore while using https:
- [x] Add macosx keychain support
- [x] Add windows keystore support

Closes #33, #108

### DIFF
--- a/lib/GLPI/Agent/HTTP/Client.pm
+++ b/lib/GLPI/Agent/HTTP/Client.pm
@@ -351,6 +351,10 @@ sub _KeyChain_or_KeyStore_Export {
                 command => "certutil -Store -Silent -Split",
                 logger  => $logger
             );
+            getAllLines(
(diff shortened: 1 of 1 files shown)


J-C-P:
ça, c'est cool !
Merci :)

Issue #35, opened 2021-12-14
Remote Management - Supremo


ddemetris:
### Documentation

Yes, I read it

### Is your feature request related to a problem?

Hello,

If it's possible to support Supremo Remote Desktop, as it's almost the same as TeamViewer

HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Supremo
    ClientID    REG_DWORD    0xXXXXXXXX

HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\teamviewer
    ClientID    REG_DWORD    0xXXXXXXXX

### Expected behavior

The Supremo ID to be shown in the Remote Management category of the computers

### Alternatives you've considered

N/A

g-bougard:
Hi @ddemetris 

this could be interesting as this product seems also to have a client for linux and MacOSX, thank you for your suggestion.

Do you know if the client installs a link binding on the system so we may also think to add such support on GLPI side ?

ddemetris:
Hello,

I have asked their support and they said that is not possible to generate a link to directly connect.

The way it works from their web based client is by using
`supremo:XXXXXXXXX`
from a browser that sends the ID to the Supremo client (tested on Windows) whether:

`HKEY_CLASSES_ROOT\supremo
    (Default)    REG_SZ    Supremo URI
    URL Protocol    REG_SZ
    Content Type    REG_SZ    application/x-supremo

HKEY_CLASSES_ROOT\supremo\shell
    (Default)    REG_SZ    open

HKEY_CLASSES_ROOT\supremo\shell\open\command
    (Default)    REG_SZ    "C:\Program Files (x86)\Supremo\Supremo.exe" "%1"
`

g-bougard:
Closing as #179 is merged.

PR #179, opened 2022-07-13
Add SupRemo in Remote_Mgmt

dxdemetriou:
Ref: https://github.com/glpi-project/glpi-agent/issues/35

### DIFF
--- a/lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/SupRemo.pm
+++ b/lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/SupRemo.pm
@@ -47,10 +47,8 @@ sub doInventory {
     my $inventory = $params{inventory};
     my $logger    = $params{logger};
 
-    my $supRemoID = _getID(
-        osname  => OSNAME,
-        logger  => $logger
-    );
+    my $supRemoID = OSNAME eq 'MSWin32' ?
+        _getID_MSWin32(logger  => $logger) : _getID_supremo_info(logger  => $logger);
     if (defined($supRemoID)) {
         $logger->debug('Found SupRemoID : ' . $supRemoID) if ($logger);
 
--- a/lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/SupRemo.pm
+++ b/lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/SupRemo.pm
@@ -19,7 +19,7 @@ sub isEnabled {
         # Depending on the installation the supremo key can be in two place in X64 OS
 
         my $key = getRegistryKey(
-            path => "HKEY_LOCAL_MACHINE/SOFTWARE/Wow6432Node/Supremo",
+            path => "HKEY_LOCAL_MACHINE/SOFTWARE/Supremo",
             # Important for remote inventory optimization
             required    => [ qw/ClientID/ ],
             maxdepth    => 1,
--- a/lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/SupRemo.pm
+++ b/lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/SupRemo.pm
@@ -90,6 +90,7 @@ sub _getID_MSWin32 {
                 path => "HKEY_LOCAL_MACHINE/SOFTWARE/Wow6432Node/Supremo",
                 # Important for remote inventory optimization
                 required    => [ qw/ClientID/ ],
+                maxdepth    => 2,
             );
         }
 
--- a/lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/SupRemo.pm
+++ b/lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/SupRemo.pm
@@ -80,7 +80,7 @@ sub _getID_MSWin32 {
 
     unless ($clientid) {
         my $supremover_reg = getRegistryKey(
-            path => "HKEY_LOCAL_MACHINE/SOFTWARE/Wow6432Node/Supremo",
+            path => "HKEY_LOCAL_MACHINE/SOFTWARE/Supremo",
             # Important for remote inventory optimization
             required    => [ qw/ClientID/ ],
             maxdepth    => 2,
--- a/lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/SupRemo.pm
+++ b/lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/SupRemo.pm
@@ -113,7 +113,7 @@ sub _getID_MSWin32 {
         }
     }
 
-    return hex2dec($clientid);
+    return sprintf("%09d",hex2dec($clientid));
 }
 
 sub _getID_supremo_info {
--- a/lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/SupRemo.pm
+++ b/lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/SupRemo.pm
@@ -70,7 +70,7 @@ sub _getID_MSWin32 {
     GLPI::Agent::Tools::Win32->use();
 
     my $clientid = getRegistryValue(
-        path => "HKEY_LOCAL_MACHINE/SOFTWARE/Supremo/Wow6432Node/ClientID",
+        path => "HKEY_LOCAL_MACHINE/SOFTWARE/Supremo/ClientID",
     );
     if (!$clientid && is64bit()) {
         $clientid = getRegistryValue(
--- a/lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/SupRemo.pm
+++ b/lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/SupRemo.pm
@@ -1,0 +1,129 @@
+package GLPI::Agent::Task::Inventory::Generic::Remote_Mgmt::SupRemo;
+
+use strict;
+use warnings;
+
+use parent 'GLPI::Agent::Task::Inventory::Module';
+
+use English qw(-no_match_vars);
+
+use GLPI::Agent::Tools;
+
+sub isEnabled {
+    my (%params) = @_;
+
+    if (OSNAME eq 'MSWin32') {
+
+        GLPI::Agent::Tools::Win32->use();
+
+        # Depending on the installation the supremo key can be in two place in X64 OS
+
+        my $key = getRegistryKey(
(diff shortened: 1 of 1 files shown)


lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/SupRemo.pm
```diff
@@ -0,0 +1,129 @@
...
+    my (%params) = @_;
+
+    if (OSNAME eq 'MSWin32') {
+
+        GLPI::Agent::Tools::Win32->use();
+
+        # Depending on the installation the supremo key can be in two place in X64 OS
+
+        my $key = getRegistryKey(
+            path => "HKEY_LOCAL_MACHINE/SOFTWARE/Wow6432Node/Supremo",
```
g-bougard: It seems you're requesting the same key than on line 30. Do you forgot to remove one `Wow6432Node` from the key path here ?
```suggestion
            path => "HKEY_LOCAL_MACHINE/SOFTWARE/Supremo",
```

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/SupRemo.pm
```diff
@@ -0,0 +1,129 @@
...
+
+    return _getID_supremo_info(%params);
+}
+
+sub _getID_MSWin32 {
+
+    GLPI::Agent::Tools::Win32->use();
+
+    my $clientid = getRegistryValue(
+        path => "HKEY_LOCAL_MACHINE/SOFTWARE/Supremo/Wow6432Node/ClientID",
```
g-bougard: As above.
```suggestion
        path => "HKEY_LOCAL_MACHINE/SOFTWARE/Supremo/ClientID",
```

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/SupRemo.pm
```diff
@@ -0,0 +1,129 @@
...
+    );
+    if (!$clientid && is64bit()) {
+        $clientid = getRegistryValue(
+            path => "HKEY_LOCAL_MACHINE/SOFTWARE/Wow6432Node/Supremo/ClientID",
+        );
+    }
+
+    unless ($clientid) {
+        my $supremover_reg = getRegistryKey(
+            path => "HKEY_LOCAL_MACHINE/SOFTWARE/Wow6432Node/Supremo",
```
g-bougard: As above
```suggestion
            path => "HKEY_LOCAL_MACHINE/SOFTWARE/Supremo",
```

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/SupRemo.pm
```diff
@@ -0,0 +1,129 @@
...
+sub doInventory {
+    my (%params) = @_;
+
+    my $inventory = $params{inventory};
+    my $logger    = $params{logger};
+
+    my $supRemoID = _getID(
+        osname  => OSNAME,
+        logger  => $logger
+    );
```
g-bougard: Don't pass osname here but directly use OSNAME to call right sub
```suggestion
    my $supRemoID = OSNAME eq 'MSWin32' ?
        _getID_MSWin32(logger  => $logger) : _getID_supremo_info(logger  => $logger);
```

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/SupRemo.pm
```diff
@@ -0,0 +1,129 @@
...
+
+sub _getID {
+    my (%params) = @_;
+
+    my $osname = delete $params{osname} // '';
+
+    return _getID_MSWin32() if $osname eq "MSWin32";
+
+    return _getID_supremo_info(%params);
+}
```
g-bougard: Remove this with the simpler code I propose to call the right sub
```suggestion
```

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/SupRemo.pm
```diff
@@ -0,0 +1,129 @@
...
+            path => "HKEY_LOCAL_MACHINE/SOFTWARE/Wow6432Node/Supremo/ClientID",
+        );
+    }
+
+    unless ($clientid) {
+        my $supremover_reg = getRegistryKey(
+            path => "HKEY_LOCAL_MACHINE/SOFTWARE/Wow6432Node/Supremo",
+            # Important for remote inventory optimization
+            required    => [ qw/ClientID/ ],
+        );
```
g-bougard: You should add maxdepth here
```suggestion
            required    => [ qw/ClientID/ ],
            maxdepth    => 2,
        );
```

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/SupRemo.pm
```diff
@@ -0,0 +1,129 @@
...
+        if(!$supremover_reg && is64bit()){
+            $supremover_reg = getRegistryKey(
+                path => "HKEY_LOCAL_MACHINE/SOFTWARE/Wow6432Node/Supremo",
+                # Important for remote inventory optimization
+                required    => [ qw/ClientID/ ],
+            );
```
g-bougard: You should add maxdepth here
```suggestion
                required    => [ qw/ClientID/ ],
                maxdepth    => 2,
            );
```

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/SupRemo.pm
```diff
@@ -0,0 +1,129 @@
...
+        return unless $supremover_reg;
+
+        # Look for subkey beginning with Version
+        foreach my $key (keys(%{$supremover_reg})) {
+            next unless $key =~ /^Version\d+\//;
+            $clientid = $supremover_reg->{$key}->{"/ClientID"};
+            last if (defined($clientid));
+        }
+    }
+
```
g-bougard, line 106: Here you must handle correctly the case clientid is not found to avoid an error on sprintf() call and just ignore the inventory.
```suggestion

    return unless $clientid;

```

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/SupRemo.pm
```diff
@@ -0,0 +1,129 @@
...
+    return sprintf("%09d",hex2dec($clientid));
+}
+
+sub _getID_supremo_info {
+    my (%params) = @_;
+
+    return getFirstMatch(
+        command => "supremo --info",
+        pattern => qr/SupRemo ID:(?:\033\[0m|\s)*(\d+)\s+/,
```
g-bougard, line 117: To avoid special chars in the regexp, can you check if there's an option for `supremo` command to have a machine readable output or just escape the pretty formatting ?

g-bougard (changes requested):
Hi @dxdemetriou 

I think you made few mistakes with the registry path. Can you check my comments and fix them ?

I also suggest you few other changes.

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/SupRemo.pm
```diff
@@ -0,0 +1,129 @@
...
+        }
+    }
+
+    return sprintf("%09d",hex2dec($clientid));
+}
+
+sub _getID_supremo_info {
+    my (%params) = @_;
+
+    return getFirstMatch(
```
g-bougard, line 115: Another point here, just to be sure: Is the clientid formatted as expected here like a number on 9 digits prefixed by zeros as the returned value of _getID_MSWin32() is suggesting it must be ?

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/SupRemo.pm
dxdemetriou: It's needed both on 32bit and 64bit the Wow6432Node. I just left both architectures on the code in case of future updates.

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/SupRemo.pm
dxdemetriou: Yes, if the prefixed zeroes are missing the sofware does not load the number and gives an error

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/SupRemo.pm
g-bougard: If you install 32bit version on 32bit OS or 64bit version on 64bit OS, the keypath should probably be `HKEY_LOCAL_MACHINE/SOFTWARE/Supremo`, and if you install 32bit version on 64bit OS, the keypath would be `HKEY_LOCAL_MACHINE/SOFTWARE/Wow6432Node/Supremo`. `Wow6432Node` node doesn't exist on 32bit OS.

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/SupRemo.pm
g-bougard: This is still wrong as on 64bit OS, if the value doesn't exist, you'll retry the same key at l.30.
I don't know Supremo, but can I guess they only provide at the moment a 32bit version ? If yes, just suppose they will provide a 64bit version in the future. And then you need to fix this code.

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/SupRemo.pm
g-bougard: Same remark as above.

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/SupRemo.pm
dxdemetriou: There is no option from the executable to have a readable output

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/SupRemo.pm
dxdemetriou: Yes, sorry, for some reason I thought the opposite :confused: . I have fixed them, thanks.

g-bougard (approved):

g-bougard:
Thank you for your contribution ;-)

PR #217, opened 2022-09-22
pci videos detect: always enabled, include 3d controller

biji:
example names:
```
VGA compatible controller Intel Corporation TigerLake-LP GT2 [Iris Xe Graphics]
3D controller NVIDIA Corporation TU117M [GeForce GTX 1650 Mobile / Max-Q]
```

### DIFF
--- a/lib/GLPI/Agent/Task/Inventory/Generic/PCI/Videos.pm
+++ b/lib/GLPI/Agent/Task/Inventory/Generic/PCI/Videos.pm
@@ -13,10 +13,7 @@ use GLPI::Agent::Tools::Generic;
 use constant    category    => "video";
 
 sub isEnabled {
-    # both windows and linux have dedicated modules
-    return
-        OSNAME ne 'MSWin32' &&
-        OSNAME ne 'linux';
+    return 1;
 }
 
 sub doInventory {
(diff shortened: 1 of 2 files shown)


lib/GLPI/Agent/Task/Inventory/Generic/PCI/Videos.pm
```diff
@@ -13,10 +13,7 @@ use GLPI::Agent::Tools::Generic;
 use constant    category    => "video";
 
 sub isEnabled {
-    # both windows and linux have dedicated modules
-    return
-        OSNAME ne 'MSWin32' &&
-        OSNAME ne 'linux';
+    return 1;
```
g-bougard, line 16: Here you shouldn't enable this test if OSNAME is `MSWin32`: getPCIDevices() call requires `lspci` command on the system and it is not available on `MSWin32`.

About linux, there's a caveat. It can generate duplicated entries as this is the case on my own system. If I run `glpi-inventory --partial video --debug --debug` I can control both `GLPI::Agent::Task::Inventory::Generic::PCI::Videos` & `GLPI::Agent::Task::Inventory::Linux::Videos` are run and the inventory reports:
```
      "videos": [
         {
            "chipset": "VGA compatible controller",
            "name": "NVIDIA Corporation TU106 [GeForce RTX 2060 SUPER]"
         },
         {
            "memory": 8192,
            "name": "NVIDIA GeForce RTX 2060 SUPER (TU106-A)",
            "resolution": "1920x1080"
         }
      ]
   },
```
If can confirm you I only have one `NVIDIA GeForce RTX 2060 SUPER` in my computer ;-)

g-bougard (changes requested):
Hi @biji,

thank you for your PR.

Anyway, it can't be merged as it is.

I think this is more complicated then it is as it can generate duplicated entries. Maybe you should better just include the regexp update on `Generic::PCI::Videos` but also use `getPCIDevices()` API lately in `Linux::Videos` to include missed videos cards.

Also, it would be great if you can include unittests.

g-bougard:
Replaced by #233

PR #218, opened 2022-09-22
use getPCIDevices() to include missed videos cards

biji:
related to https://github.com/glpi-project/glpi-agent/pull/217

### DIFF
--- a/lib/GLPI/Agent/Task/Inventory/Linux/Videos.pm
+++ b/lib/GLPI/Agent/Task/Inventory/Linux/Videos.pm
@@ -7,6 +7,7 @@ use parent 'GLPI::Agent::Task::Inventory::Module';
 
 use GLPI::Agent::Tools;
 use GLPI::Agent::Tools::Unix;
+use GLPI::Agent::Tools::Generic;
 
 use constant    category    => "video";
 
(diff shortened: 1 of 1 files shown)


lib/GLPI/Agent/Task/Inventory/Linux/Videos.pm
```diff
@@ -90,28 +91,38 @@ sub doInventory {
...
+            section => 'VIDEOS',
+            entry   => $video
+        );
 
-    if ($video->{MEMORY} && $video->{MEMORY} =~ s/kb$//i) {
-        $video->{MEMORY} = int($video->{MEMORY} / 1024);
-    }
-    if ($video->{RESOLUTION}) {
-        $video->{RESOLUTION} =~ s/@.*//;
+        $skipPci = $video->{PCIID};
```
g-bougard, line 117: By now, you should not trust $video->{PCIID} as you have no guaranty it is set.

g-bougard (changes requested):
Hi @biji 
this is better than previous PR.
Anyway you check PCIID which is not always set... sorry, this is not enough to avoid duplicate entries.

g-bougard:
Maybe linux video support should completely be rewritten. I'm still not satisfied by the current implementation as for example we can't trust datas from the Xorg log file.

biji:
using `lspci` should run on wider system, with wayland or xorg.. but can't detect memory size

g-bougard:
Replaced by #233

PR #261, opened 2022-11-19
fixed regexp in Inventory::Generic::Ipmi::Fru::Controllers

po1vo:
subj.


g-bougard:
Thank you for this fix

PR #262, opened 2022-11-20
bugfix: docker networks are virtual

po1vo:
subj.


g-bougard:
Thank you for your contribution.

PR #295, opened 2022-12-12
bugfix: ip from iproute2 underreports interfaces

po1vo:
`ip` and `ifconfig` provide different number of interfaces. Whenever a real/physical interface doesn't have an IP while an alias interface does, this real interface is masked by the alias. See the test fixtures, specifically eth0 and eth0:srv interfaces.


g-bougard:
Thank you @po1vo for this fix.

PR #299, opened 2022-12-14
Linux interfaces fix refactor

po1vo:
Found a nicer solution for my latest bugfix :0)


g-bougard:
It is indeed nicer.
In the future, if you think you'll still have some work on a PR before it could be merged, don't hesitate to prefix its title with `[wip]` and just remove it when you think you have finished ;-)

PR #347, opened 2023-02-21
implement assetname-support option

liedekef:
Related to #345 discussion.

### DIFF
--- a/Changes
+++ b/Changes
@@ -58,9 +58,10 @@ inventory:
   for a mouse as example
 * fix #344: Fix manufacturer for AG Neovo monitors
 * fix #346: Fix partition label for dirty vfat ones on linux
-* Feature: support assetname-support as option for agent and inventory
+* Feature: support assetname-support as option for agent and inventory on ux-flavors
   if 1 (the default), the short hostname is used as asset name
   if 2, the as-is hostname (can be fqdn) is used as asset name
+  Warning: this feature does not apply on MacOS or Windows
 
 remoteinventory:
 * Fix USERNAME & PASSWORD environment variable support with ssh remote inventory
--- a/Changes
+++ b/Changes
@@ -58,6 +58,9 @@ inventory:
   for a mouse as example
 * fix #344: Fix manufacturer for AG Neovo monitors
 * fix #346: Fix partition label for dirty vfat ones on linux
+* Feature: support assetname-support as option for agent and inventory
+  if 1 (the default), the short hostname is used as asset name
+  if 2, the as-is hostname (can be fqdn) is used as asset name
 
 remoteinventory:
 * Fix USERNAME & PASSWORD environment variable support with ssh remote inventory
--- a/Changes
+++ b/Changes
@@ -58,10 +58,10 @@ inventory:
   for a mouse as example
 * fix #344: Fix manufacturer for AG Neovo monitors
 * fix #346: Fix partition label for dirty vfat ones on linux
-* Feature: support assetname-support as option for agent and inventory on ux-flavors
+* Feature: support assetname-support as option for agent on most unix
   if 1 (the default), the short hostname is used as asset name
   if 2, the as-is hostname (can be fqdn) is used as asset name
-  Warning: this feature does not apply on MacOS or Windows
+  this feature does not apply on MacOS or Windows
 
 remoteinventory:
 * Fix USERNAME & PASSWORD environment variable support with ssh remote inventory
--- a/bin/glpi-agent
+++ b/bin/glpi-agent
@@ -261,6 +261,7 @@ glpi-agent [options] [--server server|--local path]
     --additional-content=FILE      additional inventory content file
     --assetname-support=1|2        if 1 (the default), the short hostname is used as asset name
                                    if 2, the as-is hostname (can be fqdn) is used as asset name
+				   Warning: this feature does not apply on MacOS or Windows
     --partial=CATEGORY             make a partial inventory of given category
                                      items, this option implies --json
     --credentials                  set credentials to support database inventory
--- a/bin/glpi-agent
+++ b/bin/glpi-agent
@@ -18,6 +18,7 @@ my $options = {};
 
 GetOptions(
     $options,
+    'assetname-support=i',
     'additional-content=s',
     'backend-collect-timeout=s',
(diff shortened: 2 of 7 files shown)


lib/GLPI/Agent/Task/Inventory/Generic/Hostname.pm
```diff
@@ -21,9 +21,15 @@ sub doInventory {
...
-    my $hostname = getHostname(short => 1);
+    my $assetname_support = $params{'assetname_support'};
+
+    # use the hostname as desired
+    my $hostname;
+    if ($assetname_support == 2 ) {
+	    $hostname = getHostname();
+    } else {
+	    $hostname = getHostname(short => 1);
+    }
```
g-bougard: On lines 29 & 31, you mixed tabulation and spaces. This is why the linux test failed. Only spaces should be used.
```suggestion
    if ($assetname_support == 2 ) {
        $hostname = getHostname();
    } else {
        $hostname = getHostname(short => 1);
    }
```

liedekef: I updated the code, let's see the tests :-)
Maybe, future-wise, vi-rules could be added to the files?

g-bougard:
You should probably also support windows here: https://github.com/glpi-project/glpi-agent/blob/b308e0942e723c737e665a985d6dceeb75a94626/lib/GLPI/Agent/Task/Inventory/Win32/Hardware.pm#L72

lib/GLPI/Agent/Task/Inventory/Generic/Hostname.pm
```diff
@@ -21,9 +21,15 @@ sub doInventory {
...
 
     my $inventory = $params{inventory};
-
-    # keep just the hostname
-    my $hostname = getHostname(short => 1);
+    my $assetname_support = $params{'assetname_support'};
+
+    # use the hostname as desired
+    my $hostname;
+    if ($assetname_support == 2 ) {
```
g-bougard: There's one space to remove here.
```suggestion
    if ($assetname_support == 2) {
```

g-bougard (changes requested):
As I still commented, you should also update the code for win32 support.

Anyway, I have another concern: assetname is not always based on hostname so this code is platform specific and so the `Changes` file update should tell which platforms are or can be concerned. Also the option description should also report the fact the option is platform specific.

As example, on MacOSX, the option wouldn't change anything for the asset name. On windows, after you'll have updated the code, there's also no guarantee the option applies. I think this can depend on the OS version...

lib/GLPI/Agent/Task/Inventory/Generic/Hostname.pm
liedekef: Concerning windows: since by default it tries to use the DNS name, I think changing that might not be desired. Also, I don't think windows has a notion of fqdn as hostname, so "short" being 0 or 1 will not change anything (and next to that: I don't have direct access to check this as admin on windows servers).
So, for now: I could put in the Changes file that this only impacts unix-flavored OSs.

liedekef: I adapted the Changes file (and others) to indicate the option does not apply on MacOS and Windows.
And fixed the spaces.
Also: in Win32/OS.pm there is also $hostname being defined, but never used ==> should I remove it there?

liedekef:
Also: in Win32/OS.pm there is also $hostname being defined, but never used ==> should I remove it there?

g-bougard:
> Also: in Win32/OS.pm there is also $hostname being defined, but never used ==> should I remove it there?

No, I also saw it this morning and still removed it, see 2da559f.

Changes
```diff
@@ -58,6 +58,10 @@ inventory:
   for a mouse as example
 * fix #344: Fix manufacturer for AG Neovo monitors
 * fix #346: Fix partition label for dirty vfat ones on linux
+* Feature: support assetname-support as option for agent and inventory on ux-flavors
+  if 1 (the default), the short hostname is used as asset name
+  if 2, the as-is hostname (can be fqdn) is used as asset name
+  Warning: this feature does not apply on MacOS or Windows
```
g-bougard: I definitively don't want to see a "Warning:" note the Changes file.
```suggestion
* Feature: support assetname-support as option for agent on most unix
  if 1 (the default), the short hostname is used as asset name
  if 2, the as-is hostname (can be fqdn) is used as asset name
  this feature does not apply on MacOS or Windows
```

bin/glpi-agent
```diff
@@ -258,6 +259,9 @@ glpi-agent [options] [--server server|--local path]
     --backend-collect-timeout=TIME timeout for inventory modules
                                      execution (30)
     --additional-content=FILE      additional inventory content file
+    --assetname-support=1|2        if 1 (the default), the short hostname is used as asset name
+                                   if 2, the as-is hostname (can be fqdn) is used as asset name
+				   Warning: this feature does not apply on MacOS or Windows
```
g-bougard: The same as in Changes file, a warning is wrong here:
```suggestion
    --assetname-support=1|2        [unix/linux only] set the asset name depending on the given value:
                                    - 1 (the default), the short hostname is used as asset name
                                    - 2, the as-is hostname (can be fqdn) is used as asset name
                                   this feature is not supported on MacOS or Windows
```

etc/agent.cfg
```diff
@@ -55,6 +55,12 @@ force = 0
 # additional inventory content file
 additional-content =
 
+# asset name returned
+# if 1 (the default), the short hostname is used as asset name
+# if 2, the as-is hostname (can be fqdn) is used as asset name
+# Warning: this feature does not apply on MacOS or Windows
```
g-bougard: The same, "warning" term is not appropriate.
```suggestion
# this feature is not supported on MacOS or Windows
```

g-bougard (changes requested):
Sorry, I don't like the "warning" term as proposed. Can you better merge my suggestions ?

bin/glpi-agent
g-bougard: Just noticed also there were few tabulations on the last line.

The same should be applied too fro the glpi-inventory script.

g-bougard (approved):

PR #381, opened 2023-04-19
Add GLPI-AgentMonitor (Windows systray Agent monitoring utility)

redddcyclone:
This is a little tool I wrote for monitoring the Agent on Windows in a quick glance. It sits on the system tray, and its icon changes color to red if the Agent has an error (i.e. not running). Left-clicking the icon displays a window that shows the Agent version, service status, startup type and current Agent status. Also, you can force an inventory directly from it by right clicking the systray icon or opening its main window and clicking "Force Inventory". 

I wrote this long ago for FusionInventory and been using it in my workplace's computers. Now I decided to share it here. It's really simple, but helpful. Should work normally for standard Agent installs.

### DIFF
--- a/Changes
+++ b/Changes
@@ -98,6 +98,8 @@ remoteinventory:
 * Updated remote modes support: perl, ssh and libssh2 for remote ssh, ssl for winrm
 * Fix libssh2 was missing from win32 packaging
 * Fix remote hostname for ssh remote inventory run from a win32 agent
+* Minor optimization by adding OSName caching to avoid recurrent same command request
+* Don't fallback on ssh if mode has been set to libssh2 only
 
 netdiscovery/netinventory:
 * Avoid to record invalid MAC Address from Netbios during netdiscovery task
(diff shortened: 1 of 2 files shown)


g-bougard (changes requested):
Hi @redddcyclone 

a big thank you for this interesting contribution.

IMHO, I would refuse it by itself as you provide an exe file which can't be trusted. I would prefer a GH Actions to build the exe from the provided sources or at least a README describing the building process so I can try to update our GH Actions to build it automatically.

You should anyway add a README to explain what licensing you want for your code or at least tell you're using the same license than GLPI-Agent.

Can you explain what's the "contrib/windows/GLPI-AgentMonitor/GLPI-AgentMonitor.aps" file ?

Are the "Thumbs.db" files really useful in a repository ?

Maybe I can try to add an option to include your contribution in the MSI installer and permit users request its installation.

redddcyclone:
Hi @g-bougard 

I will make these changes right away. I uploaded some unneeded files and didn't notice it, like the .aps (compiled resource file), the compiled executable and this Thumbs.db trash.

miguel-salles:
Please add to systray a "Open a ticket" that goes to new ticket page. This would be very usefull indeed. 
If possible I would even suggest a "Open a ticket with a screenshot" option.

PR #383, opened 2023-04-20
Add GLPI-AgentMonitor (Windows systray Agent monitoring utility)

redddcyclone:
Hi, I'm sorry for creating another pull request, but I forgot about #381 and deleted the repository after making a mess, and I don't know if it would be possible to reopen the same pull request but pulling from another repo.

Quoting the original PR,
This is a little tool I wrote for monitoring the Agent on Windows in a quick glance. It sits on the system tray, and its icon changes color to red if the Agent has an error (i.e. not running). Left-clicking the icon displays a window that shows the Agent version, service status, startup type and current Agent status. Also, you can force an inventory directly from it by right clicking the systray icon or opening its main window and clicking "Force Inventory".

I wrote this long ago for FusionInventory and been using it in my workplace's computers. Now I decided to share it here. It's really simple, but helpful. Should work normally for standard Agent installs.

Now I ignored unneeded files and added a GitHub Actions workflow for compiling the executable. You can follow it as a base for integrating it later on the MSI if wanted. For the README, I used another file in the project as base. Licensing is the same as the Agent itself.

Please comment if more changes are needed or if I made a mistake.


g-bougard:
Hi @redddcyclone 
don't worry for the previous PR, here it is a very good one.
Thank you to have followed my previous recommendation and to have included a GH Actions workflow to build it.
I'll take a deeper look at the C++ code next week.

g-bougard:
About your workflow:
1. can you add a step to upload the generated exe as artifact ?
2. About compile & link steps:
   1. can they generate more verbose output ?
   2. Or do they generate some logs you can also upload as artifact for reviewing ?

redddcyclone:
Hi @g-bougard 

I made the changes you suggested. Also, I changed the build process. Instead of calling cl, rc and link, I just call msbuild pointing to the .vcxproj file. Added building for x64 too and increased verbosity for the building process.

Now the workflow will upload artifacts for the built executables and some build-related logs for each architecture.

Please comment if I did something wrong, GitHub Actions is a new thing for me.

g-bougard:
Hi @redddcyclone 
thank you, you did a great job.
I tested successfully the built systray app. Anyway I had to install vcredist. This is not a big deal, but do you think it's possible to add a static build to not have the need to install vcredist package ?

redddcyclone:
Hi @g-bougard 

Were you testing only the x64 build? Or did you test the x86 one too? I forgot to statically link the x64 build. 
Can you test it again now?

Thanks!

g-bougard:
Indeed I only tested the x64 build. Will try with the new build of course.
Thank you for that update.

contrib/windows/GLPI-AgentMonitor/GLPI-AgentMonitor.cpp
```diff
@@ -0,0 +1,688 @@
...
+
+                if (!WinHttpQueryDataAvailable(hReq, &dwSize))
+                    break;
+                if (dwSize == 0)
+                    break;
+
+                size_t dwOffset = responseBody.size();
+                responseBody.resize(dwOffset + dwSize);
+
+                if(!WinHttpReadData(hReq, &responseBody[dwOffset], dwSize, &dwDownloaded))
```
g-bougard: I see just a readability cosmetic issue here:
```suggestion
                if (!WinHttpReadData(hReq, &responseBody[dwOffset], dwSize, &dwDownloaded))
```

contrib/windows/GLPI-AgentMonitor/GLPI-AgentMonitor.cpp
```diff
@@ -0,0 +1,688 @@
...
+    }
+}
+
+// Updates the main window statuses
+VOID CALLBACK UpdateStatus(HWND hWnd, UINT message, UINT idTimer, DWORD dwTime)
+{
+    SC_HANDLE hSc = OpenSCManager(NULL, SERVICES_ACTIVE_DATABASE, SERVICE_QUERY_STATUS);
+    SC_HANDLE hSvc = OpenService(hSc, SERVICE_NAME, SERVICE_QUERY_STATUS);
+    SERVICE_STATUS lpsvcStatus;
+    if (!QueryServiceStatus(hSvc, &lpsvcStatus)) { }
```
g-bougard: I don't see why a test to run an empty bloc would be required here. Did you construct it to eventually report a problem or just to abort in the case we don't get the service ? If yes, this may be the time to add the code, otherwise:
```suggestion
    QueryServiceStatus(hSvc, &lpsvcStatus);
```

g-bougard (commented):
Hi @redddcyclone 
I took a deeper look at the C++ code. Even if I'm definitively not a C++ expert, can you check my 2 little comments ?

redddcyclone:
Hi @g-bougard 

Gonna work on this. Some things were written as I was in the middle of other work, some years ago, and left unreviewed.

As for QueryServiceStatus, this empty if was there because Visual Studio complains I'm ignoring the return value. Actually, now is the time for some changes in the code regarding error handling. I'll work on it right now.

redddcyclone:
@g-bougard 

Just worked on your suggestions and added some error handling messages where there were none previously. Also I did some code refactoring to get rid of repeating code for the message boxes and, while on it, fixed some exceptions that I only got during debug.

On QueryServiceStatus I was doing it all wrong, so I read the documentation again and now it is fixed and error handling should be working fine.

Please comment if you see anything wrong.

g-bougard:
Hi @redddcyclone 

thank you to have updated and enhanced your code.

I just uploaded the latest produced artifact to VirusTotal and here are the results:
[GLPI-AgentMonitor-Build-x64.zip](https://www.virustotal.com/gui/file/da32019698624b57f6c5f70d3634b8497a102c928cb86ba8ab0f2c989d6fd2ca?nocache=1): No threat detected
[GLPI-AgentMonitor-Build-x86.zip](https://www.virustotal.com/gui/file/0916a30c23246083807e35762da0c36d87211911a8f1591e1c1025ed281a4178?nocache=1): 2 threats detected

Of course, this should be a false positive. I don't know if there's something to be done in code to make it appearing even more legit.

You will find more details on the exe itself there: https://www.virustotal.com/gui/file/3dabad2d4538f828a8e9943a93711cbb72d32b3b333d9222244892618e9fd374
It even shows 2 more threats detections. Looking at the [behavior pane](https://www.virustotal.com/gui/file/3dabad2d4538f828a8e9943a93711cbb72d32b3b333d9222244892618e9fd374/behavior), maybe something should be done to avoid "log keystrokes" alerts and maybe "hide graphical window" one. Strangely, the x64 exe also has a 2 threats detection: [x64 GLPI-AgentMonitor.exe](https://www.virustotal.com/gui/file/75a679b954caceac567d4b896f5c10b901a4663c28ce81e81075408b139ad0bc) and this is not reported on the artifact.

I'm still not sure if this is a good idea to include your exe in our contrib section and even in the installer if it can generate false positive on AV. Maybe you can simply create a dedicated GH project you can publish by yourself and we can promote here and in our documentation.

What do you think ?

redddcyclone:
Hi @g-bougard 

After some trial and error, I noticed that what is causing most of the false positives is statically linking the executable. If I keep it as dynamically linked, the artifacts built (.zip) will have no false positives. **But,** if I extract the x86 executable from the .zip, it gives a false positive from SecureAge, showing it just as "Malicious". The x64 executable passes just fine.

After some searching, I see people mentioning that SecureAge does report a lot of false positives, and nothing I could do fixed this. Maybe this would stop if I signed the executable but this is out of my reach.

You can see the test artifacts built at https://github.com/redddcyclone/glpi-agent/actions/runs/4926880048

Do you think it would still be bad to include the executable in the project? Sadly, in this situation, the VC runtime would have to be installed with the app as it would be dynamically linked.

g-bougard:
This is not bad, but not good.

For now, my though is you should better create a repo like `https://github.com/redddcyclone/glpi-agentmonitor` or any other appropriate name with only YOUR GLPI-Agent Monitor code. Keep it opensource with the same license so we can't blame you to use "GLPI" in the product name.

This way, your project will be a different product an the VirusTotal false positive detection can be handled separately from GLPI-Agent. We will probably fork your project in the GLPI Project so I can work on a dedicated packaging (and you'll have a PR for that ;-)) and we will be able to sign the exe and package by ourselves. And so we will be able to promote your work separately.

Would it be fine for you ?

redddcyclone:
Ok @g-bougard, I agree with you.

I did what you suggested. Repo at https://github.com/redddcyclone/glpi-agentmonitor

Should this be enough? Please comment

Thanks

g-bougard:
This looks good to me.
But it seems you choose to not build the static version. I think this is better to produce a static build.

redddcyclone:
Ah, I though it would be better as there would be less false positives, sorry. Actually, I prefer static too.

I now reverted to the static build.

g-bougard:
Okay, nice.

Having more or less false positive is not a big deal if we keep the project separated. I prefer having to handle few false positives than a dependency on a dedicated project.

Then, it's time to close this PR.

Thank you again for your nice contribution.

Just for you to know, here is a list of my next actions:
 - [x] Add a reference to your project in the [CONTRIB.md](https://github.com/glpi-project/glpi-agent/blob/develop/CONTRIB.md) file
 - [x] Fork the project in glpi-project
 - [x] Update the workflow to support code signing and release and create the related PR to your project
 - [x] Release from our fork so we can provide it signed
 - [x] Promote the released file from the doc
 - [x] Include an option in MSI to permit installation of our released file

As a side note, I'm still not sure creating a dedicated package is really useful as your exe can be easily copied by itself and asked to be started when users are logging in.

For any further question or idea, don't hesitate to open an issue in your repo pinging me or in our fork. I think I'll enable the "Discussions" section in our fork so you would be able to open any discussion there too.

PR #538, opened 2023-11-23
Feature: fqdn value for assetname support option

g-bougard:
As pointed out in Changes file:
```
* Feature: Update support assetname-support as option for agent on most unix
  if 3, the fqdn hostname is used as asset name
  this feature now works for remote ssh and for option 3, it requires perl installed
  on targeted computer and perl mode enabled on defined remote.
* Feature: Make assetname-support option also works to compute agent deviceid when
  unknown
```

IMHO this is a better answer than forcing every one to use fqdn as proposed in #535.

It also makes the option working for remoteinventory task.

### DIFF
--- a/Changes
+++ b/Changes
@@ -4,6 +4,13 @@ Revision history for GLPI agent
 
 inventory:
 * PR #531: Add SentinelOne Antivirus support on Linux, thanks to @MarcSamD
+* Feature: Update support assetname-support as option for agent on most unix
+  if 1 (the default), the short hostname is used as asset name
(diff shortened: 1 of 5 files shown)


g-bougard:
Hi @po1vo 

can you take a look to this PR as replacement candidate to #535 ?

g-bougard:
As a side note, I udpated documentation to include `assetname-support` and reflect this PR:
https://glpi-agent.readthedocs.io/en/latest/configuration.html#assetname-support

po1vo:
> can you take a look to this PR as replacement candidate to #535 ?

Looks good! Thanks.

Issue #571, opened 2023-12-27
Failure to detect memory slot size for VMware virtual machine with only one memory slot


ChristopheLacaze:
### Bug reporting acknowledgment

Yes, I read it

### Professional support

None

### Describe the bug

Memory slot size reported by glpi-agent for a Vmware virtual machine with 2 GB memory is wrong.
The value returned by the agent is 1732 (ie. hardware memory value) and not 2048 MB as returned by dmidecode

### To reproduce

Create a Vmware virtual machine with speciifcation :
- 2 Gb memory size
- 1 CPU
- Redhat family OS

execute glpi-agent

### Expected behavior

Glpi agent should report one 2048 MB size memory slot.

### Operating system

Linux

### GLPI Agent version

1.7

### GLPI version

10.0.x (See additional context below)

### GLPIInventory plugin or FusionInventory for GLPI plugin version

Not applicable

### Additional context

Here are results of dmidecode command and local inventory file generated on the virtual machine
[dmidecode-and-glpi-agent-inventory.zip](https://github.com/glpi-project/glpi-agent/files/13779330/dmidecode-and-glpi-agent-inventory.zip)

g-bougard:
Hi @ChristopheLacaze 

it seems I forgot to close this issue after I merged #572.

So I'm closing it now, but feel free to reopen if the problem is indeed not fixed.

PR #572, opened 2023-12-27
Bugfix: Memory size value for Vmware Virtual Machine

ChristopheLacaze:
Dmidecode slot memory size value overrided by /proc/meminfo value when there is only one memory slot for a Vmware virtual machine. It should not.
This commit should fix bug #571 I have created.
I haven't tested if this bug is also present other Virtual systems like Xen, VirtualBox...

### DIFF
--- a/lib/GLPI/Agent/Task/Inventory/Generic/Dmidecode/Memory.pm
+++ b/lib/GLPI/Agent/Task/Inventory/Generic/Dmidecode/Memory.pm
@@ -26,8 +26,8 @@ sub isEnabled {
 sub doInventory {
     my (%params) = @_;
 
-    my $inventory      = $params{inventory};
-    my $logger         = $params{logger};
+    my $inventory = $params{inventory};
+    my $logger    = $params{logger};
 
     my $memories = _getMemories(logger => $logger);
 
--- a/lib/GLPI/Agent/Task/Inventory/Generic/Dmidecode/Memory.pm
+++ b/lib/GLPI/Agent/Task/Inventory/Generic/Dmidecode/Memory.pm
@@ -28,7 +28,6 @@ sub doInventory {
 
     my $inventory      = $params{inventory};
     my $logger         = $params{logger};
-    my @physmemsystems = ('Physical', 'VMware');
 
     my $memories = _getMemories(logger => $logger);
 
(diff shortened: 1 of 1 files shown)


lib/GLPI/Agent/Task/Inventory/Generic/Dmidecode/Memory.pm
```diff
@@ -36,7 +37,7 @@ sub doInventory {
     # If only one component is defined and we are under a vmsystem, we can update
     # component capacity to real found size. This permits to support memory size updates.
     my $vmsystem = $inventory->getHardware('VMSYSTEM');
-    if ($vmsystem && $vmsystem ne 'Physical') {
+    if ($vmsystem && ! grep ( /^$vmsystem$/, @physmemsystems)) {
```
g-bougard: Your modification is too complex to just also ignore the VMware case in addition with the Physical one. Try the following with removing all the changes on line 29-30 & the definition of `@physmemsystems`.
```suggestion
    if ($vmsystem && $vmsystem !~ /^Physical|VMware$/) {
```
Your update finally remove the purpose of this check: permit to support memory size update. How would update memory on VMware and finally is this still detected ?

P.S.: By 'too complex', I mean you're making code harder to read.

ChristopheLacaze: Yes. I agree. This update removes the memory size update for Vmware virtual machines. As you can see in attached files given in my issue, dmidecode gives good values for VM memory settings.
There is no need to replace value given by dmidecode command with value retreived from /proc/meminfo. This value doesn't represent the real VM memory size setting. If VMware virtual machine memory size is greater than 4 GB, dmidecode command returns more than one memory slot. In that case, the memory size update check is not used and memory slot total size is correct.
In my modification, I use an array because I don't know if only these too vmsystem types are concerned by this check. I thought it was easier to add a vmsystem type in the array if needed.
If you think an "or" regex pattern is better, let's do it  😄. I think the right pattern is: ^(Physical|VMware)$
With the pattern ^Physical|VMware$, Physicalsdsd or sdsdVMware values match pattern.

g-bougard: Yes, I think the "or" regex pattern is better here. If this becomes required for another type we will just have to add it to the regex and after all there're a limited number of possible case.
Also you're right about the right pattern.
Can you update your PR ?

ChristopheLacaze: Hello,
My PR has been updated wih pattern regex check.

lib/GLPI/Agent/Task/Inventory/Generic/Dmidecode/Memory.pm
```diff
@@ -26,8 +26,8 @@ sub isEnabled {
 sub doInventory {
     my (%params) = @_;
 
-    my $inventory = $params{inventory};
-    my $logger    = $params{logger};
+    my $inventory      = $params{inventory};
+    my $logger         = $params{logger};
```
g-bougard: You should also revert this space alignment change.

ChristopheLacaze: Sorry for this omission. 😕

PR #575, opened 2024-01-03
Update Getopt.pm

JustApu:
Clarify that a "typical" installation only installs the "Computer Inventory" and "Remote Inventory" tasks. (It does not include the "ESX Remote Inventory" or "Network Inventory (SNMP)" tasks.

### DIFF
--- a/contrib/unix/installer/Getopt.pm
+++ b/contrib/unix/installer/Getopt.pm
@@ -182,7 +182,7 @@ glpi-agent-linux-installer [options]
                                      - "snap": don't install but extract snap package
     --runnow                       run agent tasks on installation (false)
     --type=INSTALL_TYPE            select type of installation (typical)
-                                     - "typical" to only install inventory task
+                                     - "typical" to only install Computer Inventory and Remote Inventory tasks
                                      - "network" to install glpi-agent and network related tasks
                                      - "all" to install all tasks
                                      - or tasks to install in a comma-separated list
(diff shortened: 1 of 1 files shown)


contrib/unix/installer/Getopt.pm
```diff
@@ -182,7 +182,7 @@ glpi-agent-linux-installer [options]
                                      - "snap": don't install but extract snap package
     --runnow                       run agent tasks on installation (false)
     --type=INSTALL_TYPE            select type of installation (typical)
-                                     - "typical" to only install inventory task
+                                     - "typical" to only install Computer Inventory and Remote Inventory tasks
```
g-bougard: There's no need to use `computer` as this is implicit. You should also avoid first letter capitalization.
```suggestion
                                     - "typical" to only install inventory and remote inventory tasks
```

JustApu: My suggested edit was based on how the info is presented on the Agent detail screens within GLPI.

My personal confusion (which you thankfully cleared up in the Q&A discussion) was thinking that "inventory" covered all the inventory tasks, not just the "Computer Inventory" task. Hoping to help the next person avoid my mistake.

Screenshot 2024-01-04 at 12 03 58

g-bougard: Okay, I understand your choice. Then can you just remove the first letter capitalization ?
```suggestion
                                     - "typical" to only install computer inventory and remote inventory tasks
```

JustApu: That's fine by me. I was just matching the text from the web UI.

g-bougard (approved):

PR #653, opened 2024-04-18
Fix accents rendering on Location

### LINKED ISSUE #503: Accents isn't being rendered on NetInventory task
### Describe the bug

We have a network Brother printer and it's Location inventory data is returning with parsing issues. Instead of "Escritório RNetworks" it's returning as seem on screenshot below.
image

### To reproduce

1. Set up on network printer the location "Escritório RNetworks".
2. Do a NetInventory task on device.

### Expected behavior

Accents should be handled as expected.

### Operating system

Other (See additional context below)

### GLPI Agent version

Other (See additional context below)

### GLPI version

Other (See additional context below)

### GLPIInventory plugin

1.3.5

### Additional context

GLPI Agent: 1.7.3
GLPI: 10.0.14

eduardomozart:
This is a safe approach, but maybe could be applied to other string fields as well. This fixes https://github.com/glpi-project/glpi-inventory-plugin/issues/503

### DIFF
--- a/lib/GLPI/Agent/SNMP/Device.pm
+++ b/lib/GLPI/Agent/SNMP/Device.pm
@@ -8,6 +8,7 @@ use GLPI::Agent::Tools::SNMP;
 use GLPI::Agent::Tools::Network;
 use GLPI::Agent::SNMP::MibSupport;
 use GLPI::Agent::SNMP::Device::Components;
+use Encode;
 
 # Supported infos are specified here:
 # http://fusioninventory.org/documentation/dev/spec/protocol/netdiscovery.html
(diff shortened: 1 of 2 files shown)


g-bougard:
Hi @eduardomozart 

I'm not convinced this is the right thing to do. It works in your case, and this is a specific case. The function you modify is used from many other places and I think you will broke some features.

Before choosing a solution, I need you provide a [snmpwalk in the expected format](https://glpi-agent.readthedocs.io/en/latest/bug-reporting.html#snmpwalk-output). The question here is first to find how this field is really encoded. The issue may still be you setup this field on the device in a wrong way.

eduardomozart:
The device location accepts accents (e.g. "ó", which is triggering the issue). Here's full snmpwalk output as requested:
[brother-hl-1210w.walk.txt](https://github.com/glpi-project/glpi-agent/files/15065030/brother-hl-1210w.walk.txt)

g-bougard:
Hi @eduardomozart 

thank you for the walk.

It helped me to reproduce the case and I just submitted a fix which is decoding the UTF-8 string in the right place: `getCanonicalString()` from `GLPI::Agent::Tools::SNMP` module.

It fixes your case and also a lot more on our private snmp walk repository.

eduardomozart:
Anyone that may find this issue: It's working as expected on GLPI-Agent 1.11 and GLPI 10.0.17.

PR #657, opened 2024-04-22
Improve Ubnt serial and SSID detection

eduardomozart:
Show SSID names instead of radio ifname on GLPI UI and fix Serial getting on UniFi AP series.

### DIFF
--- a/lib/GLPI/Agent/SNMP/MibSupport/Ubnt.pm
+++ b/lib/GLPI/Agent/SNMP/MibSupport/Ubnt.pm
@@ -34,10 +34,7 @@ our $mibSupport = [
 sub getFirmware {
     my ($self) = @_;
     
-    my $firmware = getCanonicalString($self->get(unifiApSystemVersion));
-
-    return $firmware
-        if defined($firmware);
+    return getCanonicalString($self->get(unifiApSystemVersion));
 }
 
 sub getModel {
--- a/lib/GLPI/Agent/SNMP/MibSupport/Ubnt.pm
+++ b/lib/GLPI/Agent/SNMP/MibSupport/Ubnt.pm
@@ -95,13 +95,17 @@ sub run {
                 # Defines the port alias with the name of the radio (e.g. raX)
                 $device->{PORTS}->{PORT}->{$port}->{IFALIAS} = $ifdescr;
                 # Replaces the radio port name with its respective <SSID>
-                $device->{PORTS}->{PORT}->{$port}->{IFNAME} = getCanonicalString($unifiVapEssidValues->{$index});
+                my $ifname = getCanonicalString($unifiVapEssidValues->{$index});
 
-                # raX and raiX are the network interfaces for the 2.4GHz and 5GHz radios respectively
-                if ($ifdescr =~ m/^ra\d+$/) {
-                    $device->{PORTS}->{PORT}->{$port}->{IFNAME} .= " (2.4GHz)";
-                } elsif ($ifdescr =~ m/^rai\d+$/) {
-                    $device->{PORTS}->{PORT}->{$port}->{IFNAME} .= " (5GHz)";
+                unless (empty($ifname)) {
+                    # raX and raiX are the network interfaces for the 2.4GHz and 5GHz radios respectively
+                    if ($ifdescr =~ m/^ra\d+$/) {
+                        $ifname .= " (2.4GHz)";
+                    } elsif ($ifdescr =~ m/^rai\d+$/) {
+                        $ifname .= " (5GHz)";
+                    }
+
+                    $device->{PORTS}->{PORT}->{$port}->{IFNAME} = $ifname;
                 }
 
                 last;
--- a/lib/GLPI/Agent/SNMP/MibSupport/Ubnt.pm
+++ b/lib/GLPI/Agent/SNMP/MibSupport/Ubnt.pm
@@ -83,6 +83,11 @@ sub run {
         # If you have more than one SSID there will also be more raX, raiX for each SSID.
         my $ifdescr = $device->{PORTS}->{PORT}->{$port}->{IFDESCR};
         next unless defined($ifdescr) && $ifdescr =~ /^ra/;
+		
+        # Replaces the port iftype from "Ethernet" (6) to "WiFi" (71)
+        if ($device->{PORTS}->{PORT}->{$port}->{IFTYPE} == 6) {
+            $device->{PORTS}->{PORT}->{$port}->{IFTYPE} = 71
+        }
 
         foreach my $index (keys(%$unifiVapNameValues)) {
             # Compares the device's current radio port name to the AP's radio list (e.g. raX eq raX)
@@ -97,11 +102,6 @@ sub run {
                     $device->{PORTS}->{PORT}->{$port}->{IFNAME} .= " (2.4GHz)";
                 } elsif ($ifdescr =~ m/^rai\d+$/) {
                     $device->{PORTS}->{PORT}->{$port}->{IFNAME} .= " (5GHz)";
-                }
-
-                # Replaces the port iftype from "Ethernet" (6) to "WiFi" (71)
-                if($device->{PORTS}->{PORT}->{$port}->{IFTYPE} == 6) {
-                    $device->{PORTS}->{PORT}->{$port}->{IFTYPE} = 71
                 }
 
                 last;
--- a/lib/GLPI/Agent/SNMP/MibSupport/Ubnt.pm
+++ b/lib/GLPI/Agent/SNMP/MibSupport/Ubnt.pm
@@ -33,13 +33,13 @@ our $mibSupport = [
 
 sub getFirmware {
     my ($self) = @_;
-    
+
     return getCanonicalString($self->get(unifiApSystemVersion));
 }
 
 sub getModel {
     my ($self) = @_;
-    
+
     return getCanonicalString($self->get(unifiApSystemModel));
 }
 
@@ -97,6 +97,11 @@ sub run {
                     $device->{PORTS}->{PORT}->{$port}->{IFNAME} .= " (2.4GHz)";
                 } elsif ($ifdescr =~ m/^rai\d+$/) {
                     $device->{PORTS}->{PORT}->{$port}->{IFNAME} .= " (5GHz)";
+                }
+
+                # Replaces the port iftype from "Ethernet" (6) to "WiFi" (71)
+                if($device->{PORTS}->{PORT}->{$port}->{IFTYPE} == 6) {
+                    $device->{PORTS}->{PORT}->{$port}->{IFTYPE} = 71
                 }
 
                 last;
--- a/lib/GLPI/Agent/SNMP/MibSupport/Ubnt.pm
+++ b/lib/GLPI/Agent/SNMP/MibSupport/Ubnt.pm
@@ -83,7 +83,7 @@ sub run {
         # If you have more than one SSID there will also be more raX, raiX for each SSID.
         my $ifdescr = $device->{PORTS}->{PORT}->{$port}->{IFDESCR};
         next unless defined($ifdescr) && $ifdescr =~ /^ra/;
-		
+
         # Replaces the port iftype from "Ethernet" (6) to "WiFi" (71)
         if ($device->{PORTS}->{PORT}->{$port}->{IFTYPE} == 6) {
             $device->{PORTS}->{PORT}->{$port}->{IFTYPE} = 71
--- a/lib/GLPI/Agent/SNMP/MibSupport/Ubnt.pm
+++ b/lib/GLPI/Agent/SNMP/MibSupport/Ubnt.pm
@@ -10,34 +10,44 @@ use GLPI::Agent::Tools::SNMP;
 
 # See UBNT-MIB
 
-use constant ubnt              => '.1.3.6.1.4.1.41112';
-use constant ubntWlStatApMac   => ubnt . '.1.4.5.1.4.1';
+use constant ubnt               => '.1.3.6.1.4.1.41112';
+use constant ubntWlStatApMac    => ubnt . '.1.4.5.1.4.1';
 
 # See UBNT-UniFi-MIB
 
 use constant unifiVapEssid        => ubnt . '.1.6.1.2.1.6';
 use constant unifiVapName         => ubnt . '.1.6.1.2.1.7';
 use constant unifiApSystemVersion => ubnt . '.1.6.3.6.0';
-use constant unifiApSystemModel	  => ubnt . '.1.6.3.3.0';
-
+use constant unifiApSystemModel   => ubnt . '.1.6.3.3.0';
 
 our $mibSupport = [
     {
         name    => "ubnt",
+        oid     => ubnt
+    },
+    {
+        name    => "ubnt-unifi",
         sysobjectid => getRegexpOidMatch(ubnt)
     }
 ];
 
 sub getFirmware {
     my ($self) = @_;
+    
+    my $firmware = getCanonicalString($self->get(unifiApSystemVersion));
 
-    return getCanonicalString($self->get(unifiApSystemVersion));
+    return $firmware
+        if defined($firmware);
 }
 
 sub getModel {
     my ($self) = @_;
+    
+    my $device = $self->device
+        or return;
 
-    return getCanonicalString($self->get(unifiApSystemModel));
+    return getCanonicalString($self->get(unifiApSystemModel))
+        if not defined($device->{MODEL});
 }
 
 sub getSerial {
@@ -46,7 +56,7 @@ sub getSerial {
     my $device = $self->device
         or return;
 
-    my $serial = $device->{MAC} || getCanonicalMacAddress($self->get(ubntWlStatApMac));
+    my $serial = getCanonicalMacAddress($self->get(ubntWlStatApMac)) || $device->{MAC};
     $serial =~ s/://g;
 
     return $serial;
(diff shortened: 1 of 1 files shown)


lib/GLPI/Agent/SNMP/MibSupport/Ubnt.pm
```diff
@@ -10,29 +10,90 @@ use GLPI::Agent::Tools::SNMP;
 
 # See UBNT-MIB
 
-use constant ubnt               => '.1.3.6.1.4.1.41112';
-use constant ubntWlStatApMac    => ubnt . '.1.4.5.1.4.1';
```
g-bougard: Please don't change alignment if this is not required.

lib/GLPI/Agent/SNMP/MibSupport/Ubnt.pm
```diff
@@ -10,29 +10,90 @@ use GLPI::Agent::Tools::SNMP;
...
+use constant unifiVapEssid        => ubnt . '.1.6.1.2.1.6';
+use constant unifiVapName         => ubnt . '.1.6.1.2.1.7';
+use constant unifiApSystemVersion => ubnt . '.1.6.3.6.0';
+use constant unifiApSystemModel	  => ubnt . '.1.6.3.3.0';
+
 
 our $mibSupport = [
     {
         name    => "ubnt",
-        oid     => ubnt
```
g-bougard, line 19: Here, don't modify the existing detection as you can break existing support.

Better add another detection in the list to have:
```
our $mibSupport = [
    {
        name    => "ubnt",
        oid     => ubnt
    },
    {
        name    => "ubnt-unifi",
        sysobjectid => getRegexpOidMatch(ubnt)
    }
];

File: lib/GLPI/Agent/SNMP/MibSupport/Ubnt.pm [hunk 3]
```diff
@@ -10,29 +10,90 @@ use GLPI::Agent::Tools::SNMP;
...
+    my ($self) = @_;
+
+    return getCanonicalString($self->get(unifiApSystemModel));
+}
+
 sub getSerial {
     my ($self) = @_;
 
-    my $serial = getCanonicalMacAddress($self->get(ubntWlStatApMac));
+    my $serial = $self->getMacAddress;
```
g-bougard: If you want to use cached value from getMacAddress, better look in `$device` here but keep existing code as fallback.
```suggestion
    my $device = $self->device
        or return;

    my $serial = $device->{MAC} || getCanonicalMacAddress($self->get(ubntWlStatApMac));
```

lib/GLPI/Agent/SNMP/MibSupport/Ubnt.pm
```diff
@@ -10,29 +10,90 @@ use GLPI::Agent::Tools::SNMP;
...
+	
+	my $device = $self->device
+        or return;
+	
+	my $serial = getCanonicalMacAddress($self->get(ubntWlStatApMac));
+	$serial = $device->{MAC}
+				if not defined ($serial);
+
+    return $serial;
+}
```
g-bougard: Here, you're wrong. You should not modify this method and even you're using a `$serial` variable name and just that is still wrong.

Can you explain what you're trying to do here ?

lib/GLPI/Agent/SNMP/MibSupport/Ubnt.pm
```diff
@@ -10,29 +10,90 @@ use GLPI::Agent::Tools::SNMP;
...
+				if($ifdescr =~ s/^ra(\d+)$//g) {
+					$device->{PORTS}->{PORT}->{$port}->{IFNAME} .= " (2.4GHz)";
+				} elsif($ifdescr =~ s/^rai(\d+)$//g) {
+					$device->{PORTS}->{PORT}->{$port}->{IFNAME} .= " (5GHz)";
+				}
+				
+				last;
+			}
+		}
+	}
```
g-bougard: Please don't mix tabulation and space indentation. Only use 4 spaces indentation. Also keep line empty without indentation if there is no code in it.

Keep one space after `if` statement.

Don't use `s///` in `if` condition: match test must use `m//` or the shorter `//` notation.

Can you comment your algorithm as it's hard to verify it is correct and you are mixing loops ?

You should also check `$ifdescr` is defined before using it in tests against matching test or equality test.

g-bougard (changes requested):
Hi @eduardomozart

thank you for your contribution. It seems interesting for UBnT support.

Can you check my comments and change your code accordingly ?

It would be nice if you can also provide a [full snmpwalk in the expected format](https://glpi-agent.readthedocs.io/en/latest/bug-reporting.html#snmpwalk-output) for testing.

lib/GLPI/Agent/SNMP/MibSupport/Ubnt.pm
g-bougard: Are you even sure you need to modify this method and `getSerial` one ?

lib/GLPI/Agent/SNMP/MibSupport/Ubnt.pm
```diff
@@ -13,17 +13,50 @@ use GLPI::Agent::Tools::SNMP;
...
     }
 ];
 
+sub getFirmware {
+    my ($self) = @_;
+    
+    my $firmware = getCanonicalString($self->get(unifiApSystemVersion));
+
+    return $firmware
+        if defined($firmware);
```
g-bougard: Indeed, you can simplify that code as the returned value will always been discarded if not defined, see https://github.com/glpi-project/glpi-agent/blob/fd68f16cc604b97b14cbaf59f7014e06aae4fa85/lib/GLPI/Agent/SNMP/MibSupport.pm#L151
```suggestion
    return getCanonicalString($self->get(unifiApSystemVersion));
```

lib/GLPI/Agent/SNMP/MibSupport/Ubnt.pm
```diff
@@ -13,17 +13,50 @@ use GLPI::Agent::Tools::SNMP;
...
+}
+
+sub getModel {
+    my ($self) = @_;
+    
+    my $device = $self->device
+        or return;
+
+    return getCanonicalString($self->get(unifiApSystemModel))
+        if not defined($device->{MODEL});
```
g-bougard: Here, you should use our [empty()](https://github.com/glpi-project/glpi-agent/blob/fd68f16cc604b97b14cbaf59f7014e06aae4fa85/lib/GLPI/Agent/Tools.pm#L153) function as it is also true for a defined but empty string:
```suggestion
        if empty($device->{MODEL});
```
Also `not` operator should be avoided in that context, so `!` is the right to use. Here it is included in the `empty()` function.

lib/GLPI/Agent/SNMP/MibSupport/Ubnt.pm
```diff
@@ -32,7 +65,54 @@ sub getSerial {
 sub getMacAddress {
     my ($self) = @_;
 
-    return $self->get(ubntWlStatApMac);
+    my $device = $self->device
+        or return;
+
+    return getCanonicalMacAddress($self->get(ubntWlStatApMac)) || $device->{MAC};
+}
```
g-bougard: As far I know, MAC can't be set before so I don't see why you need to modify this method. Do you really need this ? If yes, can you explain the context/case ?
IMHO, we should even not come here if the mac has still been found.

lib/GLPI/Agent/SNMP/MibSupport/Ubnt.pm
```diff
@@ -13,17 +13,50 @@ use GLPI::Agent::Tools::SNMP;
...
+
 sub getSerial {
     my ($self) = @_;
 
-    my $serial = getCanonicalMacAddress($self->get(ubntWlStatApMac));
+    my $device = $self->device
+        or return;
+
+    my $serial = getCanonicalMacAddress($self->get(ubntWlStatApMac)) || $device->{MAC};
```
g-bougard, line 52: Do you mean you encountered a case where the mac is set and ubntWlStatApMac oid remains empty ? So we should fallback on the still set mac address.

lib/GLPI/Agent/SNMP/MibSupport/Ubnt.pm
```diff
@@ -32,7 +65,54 @@ sub getSerial {
...
+             
+             # raX and raiX are the network interfaces for the 2.4GHz and 5GHz radios respectively
+             if($ifdescr =~ m/^ra(\d+)$/) {
+              $device->{PORTS}->{PORT}->{$port}->{IFNAME} .= " (2.4GHz)";
+             } elsif($ifdescr =~ m/^rai(\d+)$/) {
+              $device->{PORTS}->{PORT}->{$port}->{IFNAME} .= " (5GHz)";
+             }
+                
+             last;
+            }
```
g-bougard: Here there few things to update:
1. indentation is wrong
2. there are few lines with only space indentation, they should remain just empty
3. parenthesis are not required in the final `$ifdescr` matching tests
4. a space was missing after the `if` and `elsif` statements at the end
```suggestion
            # Compares the device's current radio port name to the AP's radio list (e.g. raX eq raX)
            if ($ifdescr eq $unifiVapNameValues->{$index}) {
                # Defines the port alias with the name of the radio (e.g. raX)
                $device->{PORTS}->{PORT}->{$port}->{IFALIAS} = $ifdescr;
                # Replaces the radio port name with its respective <SSID>
                $device->{PORTS}->{PORT}->{$port}->{IFNAME} = getCanonicalString($unifiVapEssidValues->{$index});

                # raX and raiX are the network interfaces for the 2.4GHz and 5GHz radios respectively
                if ($ifdescr =~ m/^ra\d+$/) {
                    $device->{PORTS}->{PORT}->{$port}->{IFNAME} .= " (2.4GHz)";
                } elsif ($ifdescr =~ m/^rai\d+$/) {
                    $device->{PORTS}->{PORT}->{$port}->{IFNAME} .= " (5GHz)";
                }

                last;
            }
```

lib/GLPI/Agent/SNMP/MibSupport/Ubnt.pm
```diff
@@ -32,7 +65,54 @@ sub getSerial {
...
+    my $unifiVapNameValues = $self->walk(unifiVapName) || {};
+    # The list of Radios is co-related to the list of SSIDs
+    # $unifiVapNameValues->{0} = ra0
+    # $unifiVapEssidValues->{0} = <SSID>
+
+    foreach my $port (keys(%$ports)) {
+        # For each device Radio port (raX, raiX etc.)
+        # If you have more than one SSID there will also be more raX, raiX for each SSID.
+        my $ifdescr = $device->{PORTS}->{PORT}->{$port}->{IFDESCR};
+        next unless (defined($ifdescr) && $ifdescr !~ m/^(?!ra)/);
```
g-bougard: I think your regexp is not correct here as `?!` is really specific (and didn't even know it exists before seeing it in your code). To me the real test you required is if `$ifdescr` starts by `ra`. In that case, there's a simpler notation:
```suggestion
        next unless defined($ifdescr) && $ifdescr =~ /^ra/;
```

External parenthesis are not required as `&&` operator has priority over `unless`.

g-bougard (changes requested):
Hi @eduardomozart 

thank you for the changes.

I still see few things to update.

So can you check my comments ? One is just a question you must clarify.

Also can you provide the snmpwalk I requested so I can include it in our private tests and validate this PR works as expected ?

eduardomozart:
> Hi @eduardomozart
> 
> thank you for the changes.
> 
> I still see few things to update.
> 
> So can you check my comments ? One is just a question you must clarify.
> 
> Also can you provide the snmpwalk I requested so I can include it in our private tests and validate this PR works as expected ?

Here's the snmpwalk output as you requested:

[u6-lite.walk.txt](https://github.com/glpi-project/glpi-agent/files/15076273/u6-lite.walk.txt)

Some Radios will not be reported with their respective names (will still be shown as "raX" or "raiX") as there's some Radios which are used internally for Mesh and Discovery, so there's no SSID associated to them, but user created SSID will report their names as expected. There also seems that Ubiquiti reports those ``iftypes`` as ``6`` (Ethernet) on OID but I believe they should be ``iftype`` as ``188`` (Wifi) as Aruba AP reports them as such, but GLPI doesn't import those kind of interfaces by default (I'll open ) and not sure if it would be OK to replace the OID ``iftype`` of Ubiquiti devices, but it would provide more information that the sysadmin could provide related to those radios, like "Wifi protocol version" and "Wifi mode", and also associate a "Wifi network", but GLPI doesn't seem to support importing ``188`` "iftypes" by default, so on the near future when I start working on a similar patch for Aruba devices and already reported such bug on https://github.com/glpi-project/glpi/issues/16981

Edit: I was wrong about ``get*ByMibSupport`` calls: when they return empty, the existent Device value isn't replaced by empty ones. One example would be to edit ``getMacAddress`` sub and remove the ``${device}->{MAC}`` and the device still would report MAC address as expected, so I refactor the code a little bit. Thanks for your suggestions, I'm learning a lot about Perl and code guidelines following your guidance, sorry for my mistakes as it's the first time I'm programming in Perl at all.

lib/GLPI/Agent/SNMP/MibSupport/Ubnt.pm
```diff
@@ -13,17 +13,43 @@ use GLPI::Agent::Tools::SNMP;
...
+    
+    return getCanonicalString($self->get(unifiApSystemVersion));
+}
+
+sub getModel {
+    my ($self) = @_;
+    
+    return getCanonicalString($self->get(unifiApSystemModel));
+}
+
```
g-bougard: You still have here 2 lines with only indentation as reported by tests.
```suggestion
sub getFirmware {
    my ($self) = @_;

    return getCanonicalString($self->get(unifiApSystemVersion));
}

sub getModel {
    my ($self) = @_;

    return getCanonicalString($self->get(unifiApSystemModel));
}

```

eduardomozart:
I removed the whitespaces and replaced the default iftype so the Radio interfaces are reported as WiFi as discussed in https://github.com/glpi-project/glpi/issues/16981 so the sysadmin could provide related to those radios, like "Wifi protocol version" and "Wifi mode", and also associate a "Wifi network".

lib/GLPI/Agent/SNMP/MibSupport/Ubnt.pm
```diff
@@ -32,7 +58,56 @@ sub getSerial {
...
+                if ($ifdescr =~ m/^ra\d+$/) {
+                    $device->{PORTS}->{PORT}->{$port}->{IFNAME} .= " (2.4GHz)";
+                } elsif ($ifdescr =~ m/^rai\d+$/) {
+                    $device->{PORTS}->{PORT}->{$port}->{IFNAME} .= " (5GHz)";
+                }
+
+                # Replaces the port iftype from "Ethernet" (6) to "WiFi" (71)
+                if($device->{PORTS}->{PORT}->{$port}->{IFTYPE} == 6) {
+                    $device->{PORTS}->{PORT}->{$port}->{IFTYPE} = 71
+                }
```
g-bougard: You again forgot a space after the `if` statement, not a big deal, but this is just to be coherent ;-)

Anyway why putting this here and not just after the test on line 85 ? Is description starting with "ra" not a sufficient condition to change the default type ?

lib/GLPI/Agent/SNMP/MibSupport/Ubnt.pm
```diff
@@ -32,7 +58,56 @@ sub getSerial {
...
+    foreach my $port (keys(%$ports)) {
+        # For each device Radio port (raX, raiX etc.)
+        # If you have more than one SSID there will also be more raX, raiX for each SSID.
+        my $ifdescr = $device->{PORTS}->{PORT}->{$port}->{IFDESCR};
+        next unless defined($ifdescr) && $ifdescr =~ /^ra/;
+
+        # Replaces the port iftype from "Ethernet" (6) to "WiFi" (71)
+        if ($device->{PORTS}->{PORT}->{$port}->{IFTYPE} == 6) {
+            $device->{PORTS}->{PORT}->{$port}->{IFTYPE} = 71
+        }
```
g-bougard: Just in case IFTYPE is not defined, and you missed a semicolon on line 89:
```suggestion
        if ($device->{PORTS}->{PORT}->{$port}->{IFTYPE} && $device->{PORTS}->{PORT}->{$port}->{IFTYPE} == 6) {
            $device->{PORTS}->{PORT}->{$port}->{IFTYPE} = 71;
        }
```

lib/GLPI/Agent/SNMP/MibSupport/Ubnt.pm
```diff
@@ -32,7 +58,56 @@ sub getSerial {
...
+                $device->{PORTS}->{PORT}->{$port}->{IFALIAS} = $ifdescr;
+                # Replaces the radio port name with its respective <SSID>
+                $device->{PORTS}->{PORT}->{$port}->{IFNAME} = getCanonicalString($unifiVapEssidValues->{$index});
+
+                # raX and raiX are the network interfaces for the 2.4GHz and 5GHz radios respectively
+                if ($ifdescr =~ m/^ra\d+$/) {
+                    $device->{PORTS}->{PORT}->{$port}->{IFNAME} .= " (2.4GHz)";
+                } elsif ($ifdescr =~ m/^rai\d+$/) {
+                    $device->{PORTS}->{PORT}->{$port}->{IFNAME} .= " (5GHz)";
+                }
```
g-bougard: I missed here too like in #658: you should validate IFNAME before using it.
```suggestion
                # Replaces the radio port name with its respective <SSID>
                my $ifname = getCanonicalString($unifiVapEssidValues->{$index});

                unless (empty($ifname)) {
                    # raX and raiX are the network interfaces for the 2.4GHz and 5GHz radios respectively
                    if ($ifdescr =~ m/^ra\d+$/) {
                        $ifname .= " (2.4GHz)";
                    } elsif ($ifdescr =~ m/^rai\d+$/) {
                        $ifname .= " (5GHz)";
                    }

                    $device->{PORTS}->{PORT}->{$port}->{IFNAME} = $ifname;
                }
```

g-bougard (changes requested):
Hi @eduardomozart 

sorry, I missed few needed checks yesterday. Nothing really critical, but this is just to make your code perfect.

Another point, packaging building is failing and I've included fixes in develop branch. Can you rebase your branch to include that GH Actions fixes ?

eduardomozart:
> Hi @eduardomozart
> 
> sorry, I missed few needed checks yesterday. Nothing really critical, but this is just to make your code perfect.
> 
> Another point, packaging building is failing and I've included fixes in develop branch. Can you rebase your branch to include that GH Actions fixes ?

Hello, I've applied your changes and tested them on production, they seem to be working fine. Also rebase my repo.

Issue #705, opened 2024-07-08
GLPI NetInventory - only first IP is inventoried


eduardomozart:
### Bug reporting acknowledgment

Yes, I read it

### Professional support

Yes, I know

### Describe the bug

When a device contains multiple "management" IP addresses, it seems that only the first IP (172.31.98.1) is checked when a NetInventory task is run:

<img width="976" alt="image" src="https://github.com/glpi-project/glpi-agent/assets/2974895/fa05574c-11dc-4b67-81ca-5570d04a8269">

Print of Network devices > [Device] > Network ports.

GLPI Agent Debug (Level 1) output:

```
[Mon Jul  8 16:03:39 2024][debug] new thread 12 to handle task NetInventory
[Mon Jul  8 16:03:39 2024][info] running task NetInventory
[Mon Jul  8 16:03:39 2024][debug] Current netinventory run expiration timeout: 1.0 hour
[Mon Jul  8 16:03:39 2024][debug] using 4 netinventory workers
[Mon Jul  8 16:03:40 2024][debug] #1, full snmp scan of 192.168.0.1 with credentials 2
[Mon Jul  8 16:03:40 2024][debug] #2, full snmp scan of 172.31.98.1 with credentials 2
[Mon Jul  8 16:03:41 2024][debug] #3, full snmp scan of 172.31.98.1 with credentials 2
[Mon Jul  8 16:03:41 2024][debug] #4, full snmp scan of 172.31.98.1 with credentials 2
[Mon Jul  8 16:03:55 2024][error] #1, SNMP communication error: no response from host 192.168.0.1
[Mon Jul  8 16:03:56 2024][error] #2, SNMP communication error: no response from host 172.31.98.1
[Mon Jul  8 16:03:56 2024][error] #3, SNMP communication error: no response from host 172.31.98.1
[Mon Jul  8 16:03:56 2024][error] #4, SNMP communication error: no response from host 172.31.98.1
```

Downloads:

GLPI Inventory file: [networkequipment_0_4.xml.txt](https://github.com/user-attachments/files/16134698/networkequipment_0_4.xml.txt)

### To reproduce

1. Create a new NetInventory task after a NetDiscovery task.
2. See error on GLPI Agent debug log.

### Expected behavior

If the contact with the device fails through the first IP, try the next one until the contact succeeds.

### Operating system

Windows

### GLPI Agent version

1.9

### GLPI version

10.0.16

### GLPIInventory plugin or other plugin version

[...]

PR #706, opened 2024-07-08
Remove "Magic VLAN" IP address

eduardomozart:
This is an workaround for https://github.com/glpi-project/glpi-agent/issues/705 so it removes the "Magic VLAN" IP which is not available for query.

### DIFF
--- a/lib/GLPI/Agent/SNMP/MibSupport/Aruba.pm
+++ b/lib/GLPI/Agent/SNMP/MibSupport/Aruba.pm
@@ -93,6 +93,16 @@ sub run {
     my $device = $self->device
         or return;
 
+    # Get list of device IPs
+    my $ips = $device->{IPS}->{IP};
+
+    # Remove "Magic VLAN" IP address
+    foreach my $ip (@$ips) {
+        if ($ip =~ m/^172\.31\.98/) {
+            splice(@$ips,$ip,1);
+        }
+    }
+
     # Get list of device ports (e.g. radioX_ssid_idY)
     my $ports = $device->{PORTS}->{PORT};
 


g-bougard:
Hi @eduardomozart 

thank you for your efforts,

but as I explained in #705, the public can't be handled this way and even can't be handled in glpi-agent.

eduardomozart:
Hello @g-bougard,
Even if this is an workaround for #705, this fix still can and should be applied.
This is a "Magic VLAN" IP, which is an internal IP and unavailable outside AP itself, so it can be safely excluded as it can't be queried or use this info in any useful way.

g-bougard:
I understand your purpose and I agree these ips could be removed from Aruba devices if these ips are internal.

Have you any documentation link that explains these kind of ip is internal and it has no sens to be in an inventory ?

My first concern here is you decided to filter out the ip because it makes the netinventory task fails as explained in #705, not because they should not be inventoried. The problem reported in #705 as I explained myself is more general and this PR fixing #705 is finally just a side-effect.

eduardomozart:
Hello @g-bougard, I understand your concerns.
This is documented on page 44 of "Aruba WLANs 101 and design fundamentals" at https://higherlogicdownload.s3.amazonaws.com/HPE/MigratedAssets/Aruba%20WLANs%20101%20and%20design%20fundamentals%20small.pdf
This is an internal DHCP IP range used for clients behind IAP NAT (the client IP on the firewall is shown as the VLAN IP of the IAP, in the case of #705, it's the VLAN 1 that is on 192.168.0.0/24 range).
If this IP was available, I would create an ACL to allow inter-VLAN communication between my server with GLPI Agent and the 172.31.98.X range or dig into this issue further (I created an issue at https://github.com/glpi-project/glpi-inventory-plugin/issues/523 for this), but it's not possible because this IP range (172.31.98.X) doesn't exists outside the IAP itself.

g-bougard:
This document is not really a documentation but more a showing. Anyway, 172.31.98.1 seems to me to be a real ip and you'll find all the client in “Magic VLAN”, as I see a client IP 172.31.98.42 in the schema. 172.31.98.1 seems reachable from any IAP client.

So to be honest, I don't think this ip should be removed.

eduardomozart:
This IP is a NAT one, not available outside. Outside devices (e.g. Firewall) see the management IP of the AP (192.168.0.0/24 in my case) instead of the client IP address, as it's only used for NAT. Those clients are Internal AP assigned on their WLANs, and unavailable outside.
Here's another AP in my network:
image
As you can see, this IP is used internally by each AP for NAT purposes.

g-bougard:
Yes, but it is not internal to the device, it is internal to the wifi network: this is a real ip from the point of view of AP users.

eduardomozart:
> Yes, but it is not internal to the device, it is internal to the wifi network: this is a real ip from the point of view of AP users.

This is a "Magic VLAN" IP range used for internal DHCP and used only for guests/NAT when "Virtual Controller assigned" IP mode is set instead of using "Network assigned". It doesn't seems to be possible to create a route to reach these clients because every AP uses 172.31.98.1 which translates to IAP VC IP in an IAP cluster.

PR #725, opened 2024-07-31
Prevent certificates overwriting during export

lukaszgi:
Certificates exported using `certutil` have a specific file name format, independent of their content. During exports from different stores, files are overwritten because they share the same names. I propose creating separate folders for each store. The conversion between certificate formats will search for files in nested folders.

### DIFF
--- a/lib/GLPI/Agent/HTTP/Client.pm
+++ b/lib/GLPI/Agent/HTTP/Client.pm
@@ -632,8 +632,9 @@ sub _KeyChain_or_KeyStore_Export {
             $logger->debug2("Changing to '$certdir' temporary folder");
             chdir $certdir;
 
-            while (my ($i, $command) = each @certCommands) {
-                my $storeDirname = "$i";
+            foreach my $command (@certCommands) {
+                my ($kind, $store) = $command =~ /-Split( -\w+)? -Store (\w+)$/;
+                my $storeDirname = $kind && $kind =~ /^ -(\w+)$/ ? "$1-$store" : $store;
                 mkdir $storeDirname;
                 chdir $storeDirname;
                 getAllLines(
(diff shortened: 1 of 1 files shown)


lib/GLPI/Agent/HTTP/Client.pm
```diff
@@ -619,46 +619,35 @@ sub _KeyChain_or_KeyStore_Export {
...
+                my $storeDirname = "$i";
+                mkdir $storeDirname;
+                chdir $storeDirname;
+                getAllLines(
+                    command => $command,
+                    logger  => $logger
+                );
+                chdir "..";
+            }
+
```
g-bougard: This each syntax is only supported since perl 5.12 as pointed out by this doc: https://perldoc.perl.org/functions/each

This will work perfectly on windows as we publish glpi-agent with perl 5.38 on this platform. But this means if some one tries to run it on another OS with really old perl version, this will just fail with a syntax error and the problem is this module is used on all OS.

So can I suggest you to modify this part with a more legacy supporting code:
```suggestion
            foreach my $command (@certCommands) {
                my ($kind, $store) = $command =~ /-Split( -\w+)? -Store (\w+)$/;
                my $storeDirname = $kind && $kind =~ /^ -(\w+)$/ ? "$1-$store" : $store;
                mkdir $storeDirname;
                chdir $storeDirname;
                getAllLines(
                    command => $command,
                    logger  => $logger
                );
                chdir "..";
            }

```
Of course, you should test it works as expected ;-)

g-bougard (changes requested):
Hello @lukaszgi 

thank you for this PR, it seems you found an annoying bug and find your way to fix it ! Good job.

I'm only requesting a little change to make your code support older versions of perl as this module is used on all OS flavours.

lukaszgi:
Hi @g-bougard 
I commit changes proposed by you. I dont know perl so i prepared my fix basing on tutorials.
After that change I tested glpi-agent and works correctly.

g-bougard (approved):

PR #739, opened 2024-08-20
Add SentinelOne EPP detection for macOS

mrb-x:
For now the agent report the Version,state of Protection and a value of staticSignatures in base_version.

Has  @g-bougard request ([here](https://github.com/glpi-project/glpi-agent/discussions/367)), I can't find any official and public documentation :-(

Adding 2 responses file from sentinelctl status and version in resources/macos/antivirus

### DIFF
--- a/lib/GLPI/Agent/Task/Inventory/MacOS/AntiVirus/SentinelOne.pm
+++ b/lib/GLPI/Agent/Task/Inventory/MacOS/AntiVirus/SentinelOne.pm
@@ -1,0 +1,71 @@
+package GLPI::Agent::Task::Inventory::MacOS::AntiVirus::SentinelOne;
+
+use strict;
+use warnings;
+
+use parent 'GLPI::Agent::Task::Inventory::Module';
+
+use GLPI::Agent::Tools;
+
+my $command = '/Library/Sentinel/sentinel-agent.bundle/Contents/MacOS/sentinelctl';
+
+sub isEnabled {
+    return canRun($command);
+}
+
+sub doInventory {
+    my (%params) = @_;
+
+    my $inventory = $params{inventory};
+    my $logger    = $params{logger};
+
+    my $antivirus = _getSentinelOne(logger => $logger);
+    if ($antivirus) {
+        $inventory->addEntry(
+            section => 'ANTIVIRUS',
+            entry   => $antivirus
+        );
+
+        $logger->debug2("Added $antivirus->{NAME} ".($antivirus->{VERSION}? " v$antivirus->{VERSION}":""))
+            if $logger;
+    }
+}
+
+sub _getSentinelOne {
+    my (%params) = @_;
+
+    my $antivirus = {
+        COMPANY     => "Sentinel Labs Inc.",
+        NAME        => "SentinelOne EPP",
+        ENABLED     => 0,
+    };
+
+    # Support file case for unittests if basefile is provided
+    if (empty($params{basefile})) {
+        $params{command} = "\"$command\" version";
+    } else {
+        $params{file} = $params{basefile}."-version";
+    }
+    my $version = getFirstMatch(
+        pattern => qr/^SentinelOne.* ([0-9.]+)$/,
+        %params
+    );
+    $antivirus->{VERSION} = $version if $version;
+
+    # Support file case for unittests if basefile is provided
+    if (empty($params{basefile})) {
+        $params{command} = "\"$command\" status";
+    } else {
+        $params{file} = $params{basefile}."-status";
+    }
+    my @status_lines = getAllLines(%params);
+    
+    my ($staticSignatures) = grep { /^\s.*staticSignatures:/i } @status_lines;
+    $antivirus->{BASE_VERSION} = $1
+        if $staticSignatures && $staticSignatures =~ /^\s.*staticSignatures:\s.*(\([0-9]+\))/i {
+            $antivirus->{ENABLED} = 1 if grep { /^\s.*Protection.*enabled$/i } @status_lines;
+        }
+    return $antivirus;
+}
+
+1;
--- a/lib/GLPI/Agent/Task/Inventory/MacOS/AntiVirus/SentinelOne.pm
+++ b/lib/GLPI/Agent/Task/Inventory/MacOS/AntiVirus/SentinelOne.pm
@@ -59,10 +59,6 @@ sub _getSentinelOne {
         $params{file} = $params{basefile}."-status";
     }
     my @lines = getAllLines(%params);
-    my $status = getFirstMatch(
-        pattern => qr/^\s.*Protection.*(enabled)$/i,
-        %params
-    );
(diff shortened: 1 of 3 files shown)


mrb-x:
I'm not able to isolate the staticSignature from SentinelCtl status, with the previous code.
So I fall back to my previous code, with just one command request.

lib/GLPI/Agent/Task/Inventory/MacOS/AntiVirus/SentinelOne.pm
```diff
@@ -0,0 +1,76 @@
...
+        pattern => qr/^\s.*staticSignatures:\s.*(\([0-9]+\))/i,
+        %params
+    );
+    $antivirus->{BASE_VERSION} = $base_version if $base_version;
+
+    my $status = getFirstMatch(
+        pattern => qr/^\s.*Protection.*(enabled)$/i,
+        %params
+    );
+    $antivirus->{ENABLED} = 1 if $status && $status =~ /^enabled$/i;
```
g-bougard: Firstly, not a big deal here, but you code involves to start the command 2 times with the same argument. This should be optimized.

Secondly, it seems to me the used regexp patterns are not correct. And even the test is overkill on the "ENABLED" check. For the first pattern, I think it should be better this one: `qr/^\s+staticSignatures:.*(\([0-9]+\))/i`. For the second, you should only use a match on: `/^\s+Protection:\s+(enabled)$/i`.

This can give something like:
```suggestion
    my @lines = getAllLines(%params);
    my $base_version_line = first { /^\s+staticSignatures:\s/i } @lines;
    $antivirus->{BASE_VERSION} = $1
        if $base_version_line && $base_version_line =~ /^\s+staticSignatures:.*(\([0-9]+\))/i;

    my $status = getFirstMatch(
        pattern => qr/^\s.*Protection.*(enabled)$/i,
        %params
    );
    $antivirus->{ENABLED} = 1 if first { /^\s+Protection:\s+enabled$/i } @lines;
```

g-bougard (changes requested):
Hi @mrb-x 

thank you for your nice proposition.

I see there's a little optimization to make and regexp can be a little updated. See my inline comment.

Also, I see the 2 test files, thank you to have added them.
But I don't the test file by itself. Did you miss to include it or don't you see how to use the test files ?

g-bougard:
Also a test is failing as you leaved white spaces at the end of at least one line.

mrb-x:
> I see there's a little optimization to make and regexp can be a little updated. See my inline comment.
> 
Thanks you for your perfectly working proposition. :-)

> Also, I see the 2 test files, thank you to have added them. But I don't the test file by itself. Did you miss to include it or don't you see how to use the test files ?

Hi @g-boudard,
Effectively, I don't know how to use test files.
This is my first PR, thanks to  @bristow, and my Perl was very rusted !

Thanks for your help.

g-bougard:
Trailing space removing is only relevant on agent code and so tests files can and should keep them: Test files MUST reflect exactly what a command is producing.

So I don't see why you committed such a change on a test file. I suggest you to simply revert it.

For the tests, this is not a big deal, I'll be able to quickly add it after the merge.

g-bougard:
Just check another point: SentinelOne support was still added for linux, see https://github.com/glpi-project/glpi-agent/blob/develop/lib/GLPI/Agent/Task/Inventory/Linux/AntiVirus/Sentinelone.pm

It seems the same command is used but using different options. Can you check if you can have more valuable datas with the same kind of options ?
To me, it seems "BASE_VERSION" won't look the same between Linux & MacOS.

mrb-x:
Ok, I try to revert my change on Status file.

The sentinelctl command are differents betwen macOS and Linux. For exemple no engines option on macOS, or no only status command on Linux. 
For now, I looking for details about responses from status command on macOS.
I can send you a list of macOS sentinelctl command available if needed.

g-bougard:
Okay, that's weird to have same command with so different comportment. Anyway this is up to the software editor.

In the [linux test file](https://github.com/glpi-project/glpi-agent/blob/develop/t/tasks/inventory/linux/antivirus/sentinelone.t) I see BASE_VERSION is looking more like a version: `30.5.6.5`
And the extracted one from your status output would be `120240430133940`.
They looks really different and this suggests me something is wrong in your extract or maybe in the linux module.
If you confirm `120240430133940` is correct as base version, let's play with it.

In status output, there's also an "agent" version which looks different than the one provided by version output. Are you sure this is not the current database version ?

If you want to try, you can update this file to include the test, but I'll do for you if you prefer: https://github.com/glpi-project/glpi-agent/blob/develop/t/tasks/inventory/macos/antivirus.t

bristow:
I don't know why I got into the conversation but I'm willing to get out ;) Thanks :pray:

g-bougard:
> I don't know why I got into the conversation but I'm willing to get out ;) Thanks 🙏

Lol, @mrb-x just thanked you as you probably inspired him.

You can probably click on "Unsubscribe" in the "Notifications" section on the right bar to no more be notified ;-)

mrb-x:
> In status output, there's also an "agent" version which looks different than the one provided by version output. Are you sure this is not the current database version ?  

I made a mistake during the data collection. I correct that. These values are equal, and are the Agent version.

  I discover that the value of staticSignatures change, so I choose to use it for BASE_VERSION, without any more reasons. 

Maybe it's better to remove this information, because I have no clue about it real meaning. 

On the screenshot below of SentinelOne User panel there is no good BASE_VERSION candidate, only dates about Agent.  
Screen_SO_GUI

g-bougard:
I think you should remove BASE_VERSION setting as it seems not coherent with other platforms BASE_VERSION implementation of SentilOne.

As far I understood, this information is not essential to manage this AV in GLPI.

If you have some contact with SentinelOne support, maybe you can request them to know if this data is available somewhere and if there's a simple method to get it. It will always be good to update this AV support later anyway.

I'll wait your cleanup of BASE_VERSION related code before merging this PR.

mrb-x:
I try to obtain this information through a colleague in charge of the SentinelOne EPP. No answer for now :-(
Still hope !

lib/GLPI/Agent/Task/Inventory/MacOS/AntiVirus/SentinelOne.pm
```diff
@@ -0,0 +1,72 @@
...
+        $params{command} = "\"$command\" status";
+    } else {
+        $params{file} = $params{basefile}."-status";
+    }
+    my @lines = getAllLines(%params);
+    my $status = getFirstMatch(
+        pattern => qr/^\s.*Protection.*(enabled)$/i,
+        %params
+    );
+    $antivirus->{ENABLED} = 1 if first { /^\s+Protection:\s+enabled$/i } @lines;
```
g-bougard: As reported by unittest, `$status` here is not used. You can just remove its related lines as the last test is doing the job on its own.
```suggestion
    $antivirus->{ENABLED} = 1 if first { /^\s+Protection:\s+enabled$/i } @lines;
```

g-bougard (changes requested):

Issue #757, opened 2024-09-13
Agents on Ubuntu trims alternate username field to 8 characters


topinet:
### Bug reporting acknowledgment

Yes, I read it

### Professional support

Yes, I know

### Describe the bug

When executing "glpi-inventory --partial user --debug", on "context:hardware" section, "lastloggeduser" filed is trimmed to 8 characters.

This field is populated on line 153 of lib/GLPI/Agent/Task/Inventory/Generic/Users.pm using "last" command, which will trim usernames longer than 8 characters, to avoid this a "-w" flag must be used.

1. 

### To reproduce

1. Install last GPLI agent for Linux:
```
$ glpi-agent --version
GLPI Agent (1.10-1)
Built by Debian
Source time: 2024-07-09 09:29
```
2. Execute inventory:
`glpi-inventory --partial user --debug`

### Expected behavior

lastloggeduser must show full username, not trimmed to 8 characters

### Operating system

Linux

### GLPI Agent version

v1.10

### GLPI version

10.0.16

### GLPIInventory plugin or other plugin version

Not applicable

PR #758, opened 2024-09-13
Update Users.pm to get full usernames from last command

topinet:
Bug reported in
https://github.com/glpi-project/glpi-agent/issues/757

### DIFF
--- a/lib/GLPI/Agent/Task/Inventory/Generic/Users.pm
+++ b/lib/GLPI/Agent/Task/Inventory/Generic/Users.pm
@@ -150,7 +150,7 @@ sub _getLoggedUsers {
 
 sub _getLastUser {
     my (%params) = (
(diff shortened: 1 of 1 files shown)


g-bougard (changes requested):
Hi @topinet 

thank you for your PR.

I can't accept it as is as this module is a "Generic" module and `-w` option is not supported by some OS `last` command.

Anyway, you can update your PR to change the comportment on line 160-161 to this code:
```perl
    my @lines = getAllLines(%params);
    unless (@lines) {
        $params{command} = 'last';
        @lines = getAllLines(%params)
            or return;
    }
```

The purpose here is to fallback on legacy code when no output is reported and this should be the case when the `-w` option is not supported by the local `last` command.

g-bougard (approved):

PR #824, opened 2025-01-17
Improvements to ssl-keystore parameter

eduardomozart:
Add SystemRootCA for macOS and support other OSes to Mozilla:CA in ssl-keystore. This also adds support for using certs from /etc/ssl/certs (Unix/Linux systems like Debian).

As discussed in https://github.com/glpi-project/glpi-agent/pull/823, this PR adds the SystemRootCA for macOS and remove the ``ssl-keystore`` limitation on other systems as they could rely at least on Mozilla CA store.

### DIFF
--- a/lib/GLPI/Agent/HTTP/Client.pm
+++ b/lib/GLPI/Agent/HTTP/Client.pm
@@ -600,13 +600,12 @@ sub _KeyChain_or_KeyStore_Export {
             SUFFIX      => ".pem",
         );
         my $file = $tmpfile->filename;
+        my $command = "security find-certificate -a -p";
+        $command .= " /System/Library/Keychains/SystemRootCertificates.keychain"
+            if $self->{ssl_keystore} =~ /^system-ssl-ca$/i;
         getAllLines(
-            command => "security find-certificate -a -p > '$file'",
-            logger  => $logger
-        );
-        getAllLines(
-            command => "security find-certificate -a -p /System/Library/Keychains/SystemRootCertificates.keychain >> '$file'",
-            logger  => $logger
+             command => "$command > '$file'",
+             logger  => $logger
         );
         push @certs, IO::Socket::SSL::Utils::PEM_file2certs($file)
             if -s $file;
--- a/lib/GLPI/Agent/HTTP/Client.pm
+++ b/lib/GLPI/Agent/HTTP/Client.pm
@@ -553,9 +553,6 @@ sub _setSSLOptions {
 sub _KeyChain_or_KeyStore_Export {
     my ($self) = @_;
 
-    # Only MacOSX and MSWin32 are supported
-    return unless $OSNAME =~ /^darwin|MSWin32$/;
-
     # But we don't need to extract anything if we still use an option to authenticate server certificate
     return if $self->{ca_cert_file} || $self->{ca_cert_dir} || (ref($self->{ssl_fingerprint}) eq 'ARRAY' && @{$self->{ssl_fingerprint}});
 
@@ -607,7 +604,11 @@ sub _KeyChain_or_KeyStore_Export {
             command => "security find-certificate -a -p > '$file'",
             logger  => $logger
         );
-        @certs = IO::Socket::SSL::Utils::PEM_file2certs($file)
+        getAllLines(
+            command => "security find-certificate -a -p /System/Library/Keychains/SystemRootCertificates.keychain >> '$file'",
+            logger  => $logger
+        );
+        push @certs, IO::Socket::SSL::Utils::PEM_file2certs($file)
             if -s $file;
     } else {
         my @certCommands;
--- a/lib/GLPI/Agent/HTTP/Client.pm
+++ b/lib/GLPI/Agent/HTTP/Client.pm
@@ -699,7 +699,7 @@ sub _KeyChain_or_KeyStore_Export {
     }
 
     # Include default CA file from Mozilla::CA if @certs is empty
-    if (!@certs && Mozilla::CA->require()) {
+    if ((!@certs || $self->{ssl_keystore} !~ /^system-ssl-ca$/i) && Mozilla::CA->require()) {
         my $cacert = Mozilla::CA::SSL_ca_file();
         push @certs, IO::Socket::SSL::Utils::PEM_file2certs($cacert)
             if -e $cacert;
--- a/lib/GLPI/Agent/HTTP/Client.pm
+++ b/lib/GLPI/Agent/HTTP/Client.pm
@@ -692,6 +692,13 @@ sub _KeyChain_or_KeyStore_Export {
         }
     }
 
+    # Like Mozilla::CA, but using certs from /etc/ssl/certs
+    if ($OSNAME !~ /^darwin|MSWin32$/) {
+        my $cacert = "/etc/ssl/certs/ca-certificates.crt";
+        push @certs, IO::Socket::SSL::Utils::PEM_file2certs($cacert)
+            if -e $cacert;
+    }
+
     # Always include default CA file from Mozilla::CA
     if (Mozilla::CA->require()) {
         my $cacert = Mozilla::CA::SSL_ca_file();
--- a/lib/GLPI/Agent/HTTP/Client.pm
+++ b/lib/GLPI/Agent/HTTP/Client.pm
@@ -694,9 +694,9 @@ sub _KeyChain_or_KeyStore_Export {
 
     # Like Mozilla::CA, but using certs from /etc/ssl/certs
     if ($OSNAME !~ /^darwin|MSWin32$/) {
-        my $cacert = "/etc/ssl/certs/ca-certificates.crt";
-        push @certs, IO::Socket::SSL::Utils::PEM_file2certs($cacert)
-            if -e $cacert;
+        my $sslcacert = "/etc/ssl/certs/ca-certificates.crt";
+        push @certs, IO::Socket::SSL::Utils::PEM_file2certs($sslcacert)
+            if -e $sslcacert;
     }
 
     # Always include default CA file from Mozilla::CA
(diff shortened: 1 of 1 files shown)


lib/GLPI/Agent/HTTP/Client.pm
```diff
@@ -553,9 +553,6 @@ sub _setSSLOptions {
 sub _KeyChain_or_KeyStore_Export {
     my ($self) = @_;
 
-    # Only MacOSX and MSWin32 are supported
-    return unless $OSNAME =~ /^darwin|MSWin32$/;
-
```
g-bougard, line 558: You probably misunderstood my purpose: Mozilla::CA library is supported by default on all platform.

"ssl-keystore" is only there to permit usage of certificates deployed in keystore or keychain as related OS provides tools to handle such deployment.

On other OS, we don't have such usage and people still have other solutions. So don't enable "ssl-keystore" support on other OS and keep this code.

lib/GLPI/Agent/HTTP/Client.pm
```diff
@@ -607,7 +604,11 @@ sub _KeyChain_or_KeyStore_Export {
             command => "security find-certificate -a -p > '$file'",
             logger  => $logger
         );
-        @certs = IO::Socket::SSL::Utils::PEM_file2certs($file)
+        getAllLines(
+            command => "security find-certificate -a -p /System/Library/Keychains/SystemRootCertificates.keychain >> '$file'",
+            logger  => $logger
+        );
+        push @certs, IO::Socket::SSL::Utils::PEM_file2certs($file)
```
g-bougard: That was not my idea.

I prefer to change the command on a given "ssl-keystore" value.

Here you force the run of 2 commands where this is not required for most people as Mozilla::CA should provide the public CA certificates.

I would prefer something like:
```#perl
my $command = "security find-certificate -a -p";
$command .= " /System/Library/Keychains/SystemRootCertificates.keychain"
    if $self->{ssl_keystore} =~ /^system-ssl-ca$/i;
getAllLines(
     command => "$command > '$file'",
     logger  => $logger
);
```

and even, on l.695, the test on Mozilla::CA should be changed to not load it if system public ca are used as this would be redundant:
```#perl
if (($OSNAME ne 'darwin' || $self->{ssl_keystore} !~ /^system-ssl-ca$/i) && Mozilla::CA->require()) {
```

lib/GLPI/Agent/HTTP/Client.pm
```diff
@@ -691,6 +692,13 @@ sub _KeyChain_or_KeyStore_Export {
         }
     }
 
+    # Like Mozilla::CA, but using certs from /etc/ssl/certs
+    if ($OSNAME !~ /^darwin|MSWin32$/) {
+        my $sslcacert = "/etc/ssl/certs/ca-certificates.crt";
+        push @certs, IO::Socket::SSL::Utils::PEM_file2certs($sslcacert)
+            if -e $sslcacert;
+    }
+
```
g-bougard, line 700: As I said on my first comment, this is not required as this is still the default.

g-bougard (changes requested):
Hi @eduardomozart 

I'm not agree to enable "ssl-keystore" for other OS as I explain below in my comments.

I would prefer also to change the behavior of system public CA loading for MacOS so it has to be clearly chosen by the user. You (and people) should consider this other command run as time consuming and this is probably not a good thing in most cases.

g-bougard:
> I would prefer also to change the behavior of system public CA loading for MacOS so it has to be clearly chosen by the user. You (and people) should consider this other command run as time consuming and this is probably not a good thing in most cases.

To complete my comment, I think this case should only be used if people are in the case glpi server certificate public CA is not available in Mozilla::CA but it is in system public CA. And this should only happen in really rare cases.

lib/GLPI/Agent/HTTP/Client.pm
eduardomozart: I believe there should be cases where internal CA is used and some automation tool already added the self-signed certificate to system CA store and it's not a public certificate provided by Mozilla::CA. As this use case is supported on Windows and macOS (otherwise it wouldn't need to import the keystore/keychain from those systems), I don't see why don't include support for keystore of other systems too, only if the SSL library loads it by default?

lib/GLPI/Agent/HTTP/Client.pm
eduardomozart: The Mozilla CA may not include self-signed CA certs available on system store. I updated the macOS code to meet your expectations, but I still believe the system root CA store should be loaded instead of the current user (root) as it's useless since it contains only Apple CA's - I know the default CA store from the user can be changed but most users won't do that.
I updated the code to load Mozilla::CA only if ``@certs`` is empty (it means it didn't load the CAs from third party cert stores) or if it's macOS and user didn't provided the 'system-root-ca' parameter.

lib/GLPI/Agent/HTTP/Client.pm
eduardomozart: From what I've seen, the Mozilla::CA isn't loaded because this line exit the function if OS is not macOS or Windows, but if Mozilla::CA is supported on Unix/Linux systems but this support isn't enable because the code exits the function before importing it.

g-bougard: Mozilla::CA is used by LWP::UserAgent http client by default.

The purpose of `_KeyChain_or_KeyStore_Export()` function is to load keystore or keychain certificates and provide them via "SSL_ca" param of IO::Socket::SSL ssl_opts() method. Using this params also means default Mozilla::CA won't be loaded by IO::Socket::SSL. So this means we need to add them ourselves to SSL_ca param.

eduardomozart: I see, thank you for your clarification.

g-bougard:
Hi @eduardomozart 

as I'm about to make a new release in the next few days, I preferred to merge manually your PR removing no more required changes.

For me, on MacOSX, the keychain default support permits people to install glpi server certificate in the safe keychain of root user. Even if most people won't do that, this is a feature they can easily use.

Now, the "ssl-keystore=system-ssl-ca" option is a security for users: Actually Mozilla::CA is updated during releases, but in the case the project dies in a far future, using this option may help people to continue using glpi-agent with server installed with SSL signed by a public CA not known at the time of glpi-agent release.

eduardomozart:
That's great!
I updated the ``ssl-keystore`` documentation to describe this new behavior, please let me know if I'm right: https://github.com/glpi-project/doc-agent/pull/11

eduardomozart:
It's a bit sad that the support for extracting CA trust store hasn't been approved for Unix/Linux and it relies only on Mozilla:CA since, as you'd said, if someday the GLPI-Agent dies and embedded Mozilla::CA becomes old, or if the user uses a self-signed CA imported to CA trust store, it will not be trusted by default.

g-bougard:
`ca-cert-file`, `ca-cert-dir` and `ssl-fingerprint` were there before `ssl-keystore` and will always be there to help people with such a problem. You can use these options to trust a certificate which is not known from the embedded Mozilla::CA.

eduardomozart:
> `ca-cert-file`, `ca-cert-dir` and `ssl-fingerprint` were there before `ssl-keystore` and will always be there to help people with such a problem. You can use these options to trust a certificate which is not known from the embedded Mozilla::CA.

Yes, I know, but it requires manual intervention, while ``ssl-keystore`` works on-the-fly. If the certificate expires (e.g. Let's Encrypt), it requires updating the signature. Also, some software antivirus like ESET intercepts the SSL connection using it own self-signed certificate to SSL inspection, and it requires creating a manual exception to exclude the GLPI server address from inspection if GLPI-Agent uses a SSL connection.

g-bougard:
I'm not agree with you. "ssl-keystore" feature would also require a manual intervention in that hypothetical case. Adding an overload charging more certificates where this is not required most of the time and for most users was not an option.
And sorry, your comment on ESET is totally out of scope.

eduardomozart:
> I'm not agree with you. "ssl-keystore" feature would also require a manual intervention in that hypothetical case. Adding an overload charging more certificates where this is not required most of the time and for most users was not an option. And sorry, your comment on ESET is totally out of scope.

It depends. If it's the Let's Encrypt CA, it's updated by updating the package ``ca-certificates`` which keep the OpenSSL CA trust store always updated without relying on GLPI-Agent updates so GLPI-Agent would keep working forever even if GLPI-Agent is long gone and built-in Mozilla::CA isn't updated. The ``ssl-fingerprint`` is useless for Let's Encrypt certificates since every 3 months the GLPI server certificate is updated, which makes it useless, and ``ca-cert-file`` and ``ca-cert-dir`` requires manual creation. I believe that ``ssl-fingerprint`` would be more useful if it applies to Root CA thumbprint since renewing the Root CA certificate doesn't affect it's thumbprint (aka. make ``ssl-fingerprint`` validates every thumbprint on the certificate chain, not only the server certificate), so even if the server certificate is changed, if the Root CA is still the same (even if renewed, since thumbprint of Root CA certificates isn't changed when it's renewed and auto-renew is enable for Windows ADCS self-signed Enterprise CA), the GLPI-Agent will continue to function forever.
Also, I'm not sure how GLPI-Agent behavior when the GLPI server-certificate expires - does it relies only on thumbprint or there's some internal SSL function which is triggered to allow/not allow connection on those cases?
About the ESET, it may also affect other products, like Sophos Firewall/FortiGate that does SSL inspection on the edge of the network, which replaces the GLPI server certificate with it's own during GLPI-Agent connection and GLPI-Agent connection fails. If Unix/Linux provides a trust store that could be extracted and used instead of relying on Mozilla::CA, I don't see why not use it, and use Mozilla::CA on last case. Some of these scenarios are not atypical at all - ESET, Kaspersky and other AV solutions (e.g. which could be installed on a endpoint of an user that works HomeOffice) does SSL interception, and the import of Windows keystore makes the GLPI-Agent works on-the-fly on Windows and macOS machines only, excluding Unix/Linux users.

g-bougard:
Hi @eduardomozart 

sorry, but PR comments thread is not a forum.

The PR has been merged manually with some changes, end of the discussion.

If you need to discuss on a point, please open a new [discussion](https://github.com/glpi-project/glpi-agent/discussions).

PR #836, opened 2025-01-24
Improve Brother SNMP printer inventory (ifType)

eduardomozart:
This commit improves the inventory of network ports of Brother printers. I have a HL-1210W printer which reports it's ifType as "``iso88023Csmacd(7)``" (Ethernet) instead of "``ieee80211(71)``" (Wireless). I'd opened a bug report with Brother but it's a common issue with other models as Brother [QL-710W](https://gist.github.com/pklaus/6fa17bb13eddd74bcfb6). It happens because Brother uses https://mibs.observium.org/mib/BROTHER-MIB/#brMultiIFConfigureTable to determine interface type. This PR queries those values and change the port type accordingly.
I also would like to know if it's possible to set the Wi-Fi Network which the printer belongs through netinventory tasks, so I could use ``$brpsWLanName->{(keys %{$brpsWLanName})[0]}`` as seen in https://github.com/glpi-project/glpi-agent/commit/ae8828bbe12b747d0bcdd3019040383a6b4de057. The only solution I could found was renaming the interface name to reflect the Wi-Fi Network name, but if the interface name changes, the interface metrics history is "lost" (unassociated) since a new interface is created in GLPI.

Brother HL-1210W snmpwalk: [snmpwalk-brother.txt](https://github.com/user-attachments/files/18558155/snmpwalk-brother.txt)

### DIFF
--- a/lib/GLPI/Agent/SNMP/MibSupport/BrotherNetConfig.pm
+++ b/lib/GLPI/Agent/SNMP/MibSupport/BrotherNetConfig.pm
@@ -101,7 +101,7 @@ sub updatePortIfType {
     foreach my $index (keys %{$brMultiIFType}) {
         foreach my $port (keys %{$ports}) {
             next unless defined($ports->{$port}->{IFNAME}) && defined($brMultiIFNodeType->{$index});
-            if ($ports->{$port}->{IFNAME} eq $brMultiIFNodeType->{$index}) {
+            if ($ports->{$port}->{IFNAME} eq getCanonicalString($brMultiIFNodeType->{$index})) {
                 # wirelesslan(2)
                 if (defined($brMultiIFType->{$index}) && isInteger($brMultiIFType->{$index}) && int($brMultiIFType->{$index}) == 2) {
                     # ieee80211(71)
--- a/lib/GLPI/Agent/SNMP/MibSupport/BrotherNetConfig.pm
+++ b/lib/GLPI/Agent/SNMP/MibSupport/BrotherNetConfig.pm
@@ -93,7 +93,7 @@ sub getPortIfType {
     # Get list of device ports types (lan(1)/wirelesslan(2))
     my $brMultiIFType = $self->walk(brMultiIFType) || {};
 
-    # Get list of device ports names 
+    # Get list of device ports names
     my $brMultiIFNodeType = $self->walk(brMultiIFNodeType) || {};
 
     foreach my $index (keys %{$brMultiIFType}) {
--- a/lib/GLPI/Agent/SNMP/MibSupport/BrotherNetConfig.pm
+++ b/lib/GLPI/Agent/SNMP/MibSupport/BrotherNetConfig.pm
@@ -19,7 +19,9 @@ use constant    printerinfomation   => net_peripheral . '.4.2.1.5.5' ;
 use constant    brInfoSerialNumber  => printerinfomation . '.1.0' ;
 use constant    brScanCountCounter  => printerinfomation . '.54.2.2.1.3.3';
 
-use constant    brpsWLanName        => brother . '.2.4.3.100.11.1.3';
+use constant    brMultiIFConfigureEntry => brother . '.2.4.4.1240.1.5.1';
+use constant    brMultiIFType           => brMultiIFConfigureEntry . '.2';
+use constant    brMultiIFNodeType       => brMultiIFConfigureEntry . '.8';
 
 # Brother NetConfig
 use constant    brnetconfig => brother . '.2.4.3.1240' ;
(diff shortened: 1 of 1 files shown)


lib/GLPI/Agent/SNMP/MibSupport/BrotherNetConfig.pm
```diff
@@ -77,6 +81,34 @@ sub getModel {
     return $model;
 }
 
+sub getPortIfType {
```
g-bougard: You should probably better rename the function to "updatePortIfType".

lib/GLPI/Agent/SNMP/MibSupport/BrotherNetConfig.pm
```diff
@@ -77,6 +81,34 @@ sub getModel {
...
+    my ($self) = @_;
+
+    my $device = $self->device
+        or return;
+
+    # Get list of device ports
+    my $ports = $device->{PORTS}->{PORT};
+
+    # Get list of device ports types (lan(1)/wirelesslan(2))
+    my $brMultiIFType = $self->walk(brMultiIFType) || {};
```
g-bougard: Here it would be simpler to just return as the result will be the same: nothing would be changed.
```suggestion
    # Get list of device ports types (lan(1)/wirelesslan(2))
    my $brMultiIFType = $self->walk(brMultiIFType)
        or return;
```

lib/GLPI/Agent/SNMP/MibSupport/BrotherNetConfig.pm
```diff
@@ -77,6 +81,34 @@ sub getModel {
...
+
+    # Get list of device ports names
+    my $brMultiIFNodeType = $self->walk(brMultiIFNodeType) || {};
```
g-bougard: Here it would be simpler to just return as the result will be the same: nothing would be changed.
```suggestion
    # Get list of device ports names
    my $brMultiIFNodeType = $self->walk(brMultiIFNodeType)
        or return;
```

lib/GLPI/Agent/SNMP/MibSupport/BrotherNetConfig.pm
```diff
@@ -77,6 +81,34 @@ sub getModel {
...
+
+    foreach my $index (keys %{$brMultiIFType}) {
+        foreach my $port (keys %{$ports}) {
+            if ($ports->{$port}->{IFNAME} eq $brMultiIFNodeType->{$index}) {
```
g-bougard: Here you have no guarantee hash fields are defined and comparing not defined values will fail with an error. So you just have to check they are defined.
```suggestion
        foreach my $port (keys %{$ports}) {
            next unless defined($ports->{$port}->{IFNAME}) && defined($brMultiIFNodeType->{$index});
            if ($ports->{$port}->{IFNAME} eq $brMultiIFNodeType->{$index}) {
```

lib/GLPI/Agent/SNMP/MibSupport/BrotherNetConfig.pm
```diff
@@ -77,6 +81,34 @@ sub getModel {
...
+                # wirelesslan(2)
+                if (defined($brMultiIFType->{$index}) && $brMultiIFType->{$index} eq 2) {
```
g-bougard: You should better check it is an integer. And there's a function for that in GLPI::Agent::SNMP::Tools. Than a numeric test would be cleaner.
```suggestion
                # wirelesslan(2)
                if (defined($brMultiIFType->{$index}) && isInteger($brMultiIFType->{$index}) && int($brMultiIFType->{$index}) == 2) {
```

g-bougard (changes requested):
Hi @eduardomozart 

I'm essentially okay with the PR.

Bellow you'll see essentially advice. B the "defined()" checks here are essentials.

Also can you attach another walk this time with the right format using the command described here: https://glpi-agent.readthedocs.io/en/nightly/bug-reporting.html#snmpwalk-output

Thank you

eduardomozart:
Hello @g-bougard, I've updated this PR with your suggestions, tested it and it's working as expected. Also updated the snmpwalk file.

lib/GLPI/Agent/SNMP/MibSupport/BrotherNetConfig.pm
```diff
@@ -77,6 +81,37 @@ sub getModel {
...
+        or return;
+
+    # Get list of device ports names
+    my $brMultiIFNodeType = $self->walk(brMultiIFNodeType)
+        or return;
+
+    foreach my $index (keys %{$brMultiIFType}) {
+        foreach my $port (keys %{$ports}) {
+            next unless defined($ports->{$port}->{IFNAME}) && defined($brMultiIFNodeType->{$index});
+            if ($ports->{$port}->{IFNAME} eq $brMultiIFNodeType->{$index}) {
```
g-bougard: Can you please update the test here to use `getCanonicalString()` ? This will guaranty I can use your snmpwalk output in our internal tests.
```suggestion
            if ($ports->{$port}->{IFNAME} eq getCanonicalString($brMultiIFNodeType->{$index})) {
```

eduardomozart:
Helo @g-bougard, tested your commit and it's working as expected.

g-bougard:
Thank you for your contribution @eduardomozart

PR #842, opened 2025-01-29
Update install-deb.sh

cbka:
add apt update before apt-get install
this is useful when using the script in e.g. ansible environments because the install script itself becomes independent from previous operations.

### DIFF
--- a/contrib/unix/install-deb.sh
+++ b/contrib/unix/install-deb.sh
@@ -61,6 +61,7 @@ downloadurlesx=$BaseUrl\glpi-agent-task-esx_$version\_all.deb
 # Setup dependencies for Agent Core
 echo "Installing agent dependencies"
 {
+    apt update
     apt-get -y install dmidecode hwdata ucf hdparm
     apt-get -y install perl libuniversal-require-perl libwww-perl libparse-edid-perl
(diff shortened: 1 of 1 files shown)


contrib/unix/install-deb.sh
```diff
@@ -92,6 +93,7 @@ echo
 if $tasknetwork; then
     echo "network task is requested"
     echo "installing dependencies"
+    apt update
```
g-bougard, line 96: I don't think this is required if still done on l.64.

contrib/unix/install-deb.sh
```diff
@@ -106,6 +108,7 @@ echo
 if $taskdeploy; then
     echo "deploy task is requested"
     echo "installing dependencies"
+    apt update
```
g-bougard, line 111: As above, I don't think this is required if still done on l.64.

g-bougard (changes requested):
Hello @cbka 

I'm okay to add one `apt update`, but I don't think this is required after the one has been run.

g-bougard:
Hi @cbka 

I finally merge manually to apply myself the patch I requested: only one `apt update` is required

Thank you for your contribution

PR #852, opened 2025-02-06
Do PROLOG and fallback to CONTACT if supported

### LINKED ISSUE #851: GLPI-Agent do not send CONTACT on first inventory
### Bug reporting acknowledgment

Yes, I read it

### Professional support

None

### Describe the bug

Hello @g-bougard, here I'm again. =D
I noticed that when using /RUNNOW, running the GLPI-Agent Monitor to force inventory the first time or installing the agent and wait the next-run (first run), the **Supported Tasks** section of the agent isn't filled, as can be seen below:

Image

I believe it happens because the https://github.com/glpi-project/glpi-agent/blob/c3fd5106b5d98db9d4e872796556d34fd4ece0cc/lib/GLPI/Agent.pm#L218 returns false which make the PROLOG call not be made and the **Supported tasks** not be reported to GLPI server on the first GLPI-Agent run.

It can become an issue because on https://github.com/glpi-project/glpi-inventory-plugin/pull/627 the GLPI-Inventory plugin also checks the **Supported tasks** reported by GLPI-Agent (now the ``use_module*`` columns of agent is checked on ``isAgentCanDo()`` PHP function call). You can see on the ``glpi-agent.log`` files (in attachment) it returns a "400" HTTP error when attempting to retrieve the "Deploy" and "Collect" tasks the first time GLPI-Agent runs and it may happen because the GLPI server do not send tasks (e.g. Deploy or Collect tasks) to agents that doesn't supports it. If the PR https://github.com/glpi-project/glpi-inventory-plugin/pull/603 be approved as it is it will not show those agents on taskjob dropdown until a second inventory task is run, which may lead the users to be confused about "the agent is shown on GLPI but not on the taskjob dropdown" kind of issues.

### To reproduce

1. Uninstall the GLPI-Agent on the client operating system (if it exists)
2. Delete agent and computer asset of the agent on GLPI-server (if it exists).
3. Install/Reinstall the GLPI-Agent.
4. Do the first inventory.
5. **Supported tasks** from GLPI isn't updated to reflect the tasks supported by GLPI-Agent installed on the client operating system.

### Expected behavior

**Supported tasks** from GLPI is updated to reflect the tasks supported by GLPI-Agent installed on the client operating system.

### Operating system

Windows

### GLPI Agent version

v1.12

### GLPI version

GLPI 10.0.18-dev (Nightly Build)

### GLPIInventory plugin or other plugin version

GLPI Inventory v1.4.0

### Additional context

I tried to change https://github.com/glpi-project/glpi-agent/blob/c3fd5106b5d98db9d4e872796556d34fd4ece0cc/lib/GLPI/Agent.pm#L218 to ``if ($target->isGlpiServer()) {`` but the GLPI server returns a 404 error and GLPI-Agent returns ``No supported answer from server at https://glpi.example.com/plugins/glpiinventory`` (``glpi-agent-1st-force-glpiserver.log``), so I believe this PR may requires changes on GLPI Core to accept those kind of requests if the Agent doesn't exists yet OR a modification be made on GLPI-Agent itself to run CONTACT again if target has been identified as ``isGlpiServer`` after inventory (so the GLPI server would accept the request and fill the **Supported Tasks** section of agent on GLPI server).

glpi-agent.log files (Debug level = 2):

[glpi-agent-1st-run.log](https://github.com/user-attachments/files/18680939/glpi-agent.log)
[glpi-agent-2nd-run.log](https://github.com/user-attachments/files/18681041/glpi-agent-2nd-run.log)

[glpi-agent-1st-next-run.log](https://github.com/user-attachments/files/18681088/glpi-agent-1st-next-run.log)

[glpi-agent-1st-force-glpiserver.log](https://github.com/user-attachments/files/18681180/glpi-agent-1st-force-glpiserver.log)

eduardomozart:
Fix #851. I just flipped the "if" to do PROLOG first and fallback to CONTACT, this makes the Supported tasks be available on GLPI server before the inventory is made the first time.

### DIFF
--- a/lib/GLPI/Agent.pm
+++ b/lib/GLPI/Agent.pm
@@ -317,7 +317,7 @@ sub runTarget {
         if (ref($response) !~ /^GLPI::Agent::Protocol::/) {
             $self->{logger}->info("$target->{id} is not understanding GLPI Agent protocol");
             $target->isGlpiServer('false');
-            return $self->runTarget($target) unless $response->expiration;
+            return 0;
         }
 
         # Handle contact answer including expiration and/or errors
@@ -401,6 +401,8 @@ sub runTarget {
 
         # Keep contact response
         $contact_response = $response;
+    } else {
+        return $self->runTarget($target) unless $response->expiration;
     }
 
     # Used when running tasks after a taskrun event
--- a/lib/GLPI/Agent.pm
+++ b/lib/GLPI/Agent.pm
@@ -401,8 +401,6 @@ sub runTarget {
 
         # Keep contact response
         $contact_response = $response;
-    } else {
-        return $self->runTarget($target) unless $response->expiration;
     }
 
     # Used when running tasks after a taskrun event
--- a/lib/GLPI/Agent.pm
+++ b/lib/GLPI/Agent.pm
@@ -210,11 +210,64 @@ sub runTarget {
         $self->{logger}->info("target $target->{id}: " . $target->getType() . " " . $target->getName());
     }
 
-    # the prolog/contact dialog must be done once for all tasks,
-    # but only for server targets
+    # By default, PROLOG request could be avoided when communicating with a GLPI server
+    # But it still may be required if we detect server supports any task due to glpiinventory plugin
     my ($response, $contact_response);
     my $client;
     my @plannedTasks = $target->plannedTasks();
+    if ($target->isType('server') && $target->doProlog()) {
+
+        return unless GLPI::Agent::HTTP::Client::OCS->require();
+
+        my $agentid;
+        # We may have to simulate a legacy PROLOG call if we just need to get an XML answer as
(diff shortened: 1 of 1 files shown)


g-bougard:
Hi @eduardomozart 

thank you for the attempt, but I will implement myself a fix, and be sure I still have some though on the right way to do it.

Actually, as this problem concerns the core part of glpi-agent and there are too many sensible parts around, I really prefer to handle it myself. Just be patient.

eduardomozart:
No problem, just wanted to help. Not sure how to handle it properly if not this way because GLPI server seems to reject any first request attempt without proper PROLOG. I believe the solution you'll find will be similar to the one in this PR. Maybe create a variable to track last inventory date and if it doesn't exists do a PROLOG first, or as you said if it's a JSON do a PROLOG, but it could cause a PROLOG be sent twice with the code as it is today.

PR #869, opened 2025-03-11
Fix Proxmox/LXC memory conversion Lxc.pm

Jo-NMS:
Issue:
The glpi-agent was incorrectly handling memory values for LXC containers on hosts using Proxmox.
Proxmox returns a raw byte value (e.g., "10737418240") without a unit suffix, and the current logic
in getCanonicalSize() immediately returns such plain numeric values. This leads to storing the
value in bytes instead of the expected megabytes (MB), potentially causing an overflow in an
unsigned int database column.

Solution:
- For values ending with a unit (K, M, G, T, or P), the code appends "b" (e.g., "16384M" becomes "16384Mb")
  so that getCanonicalSize() processes them correctly.
- For plain numeric values (e.g., "10737418240"), the code now detects them using a regex and converts
  the value from bytes to MB by dividing by (1024 * 1024).

Changes:
- Added an `elsif` branch to check if the memory value is a plain number using `/^\d+$/`.
- For these numeric values, the code converts them to MB via:
  `int($val / (1024 * 1024))`
- Retained the existing logic for values with a unit suffix to preserve backward compatibility.

### DIFF
--- a/lib/GLPI/Agent/Task/Inventory/Virtualization/Lxc.pm
+++ b/lib/GLPI/Agent/Task/Inventory/Virtualization/Lxc.pm
@@ -102,10 +102,17 @@ sub  _getVirtualMachine {
                 if $val =~ $mac_address_pattern;
         }
 
-        if ($key eq 'lxc.cgroup.memory.limit_in_bytes' || $key eq 'lxc.cgroup2.memory.max') {
-            $val .= "b" if $val =~ /[KMGTP]$/i;
-            $container->{MEMORY} = getCanonicalSize($val, 1024);
-        }
+    	if ($key eq 'lxc.cgroup.memory.limit_in_bytes' || $key eq 'lxc.cgroup2.memory.max') {
+    	    if ($val =~ /[KMGTP]$/i) {
+    		$val .= "b";
+    		$container->{MEMORY} = getCanonicalSize($val, 1024);
+    	    } elsif ($val =~ /^\d+$/) {
+    		# Plain numeric value in bytes, convert to MB
+    		$container->{MEMORY} = int($val / (1024 * 1024));
+    	    } else {
+    		$container->{MEMORY} = getCanonicalSize($val, 1024);
+    	    }
+    	}	
 
         # Update container name in Proxmox environment
         if ($proxmox && ($key eq 'lxc.uts.name' || $key eq 'lxc.utsname')) {
--- a/lib/GLPI/Agent/Task/Inventory/Virtualization/Lxc.pm
+++ b/lib/GLPI/Agent/Task/Inventory/Virtualization/Lxc.pm
@@ -102,17 +102,10 @@ sub  _getVirtualMachine {
                 if $val =~ $mac_address_pattern;
         }
 
-    	if ($key eq 'lxc.cgroup.memory.limit_in_bytes' || $key eq 'lxc.cgroup2.memory.max') {
-    	    if ($val =~ /[KMGTP]$/i) {
-    		$val .= "b";
-    		$container->{MEMORY} = getCanonicalSize($val, 1024);
-    	    } elsif ($val =~ /^\d+$/) {
-    		# Plain numeric value in bytes, convert to MB
-    		$container->{MEMORY} = int($val / (1024 * 1024));
-    	    } else {
-    		$container->{MEMORY} = getCanonicalSize($val, 1024);
-    	    }
-    	}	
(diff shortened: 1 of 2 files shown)


lib/GLPI/Agent/Task/Inventory/Virtualization/Lxc.pm
```diff
@@ -102,10 +102,17 @@ sub  _getVirtualMachine {
...
+    	    if ($val =~ /[KMGTP]$/i) {
+    		$val .= "b";
+    		$container->{MEMORY} = getCanonicalSize($val, 1024);
+    	    } elsif ($val =~ /^\d+$/) {
+    		# Plain numeric value in bytes, convert to MB
+    		$container->{MEMORY} = int($val / (1024 * 1024));
+    	    } else {
+    		$container->{MEMORY} = getCanonicalSize($val, 1024);
+    	    }
+    	}
```
g-bougard: Hi @Jo-NMS 

Looking back at getCanonicalSize API, you should probably better do the following:
```suggestion
        if ($key eq 'lxc.cgroup.memory.limit_in_bytes' || $key eq 'lxc.cgroup2.memory.max') {
            $val .= $val =~ /[KMGTP]$/i ? "b" : $val =~ /^\d+$ ? "bytes" : "";
            $container->{MEMORY} = getCanonicalSize($val, 1024);
        }
```

Also try to always use the same indentation scheme and on glpi-agent project we just use 4 spaces indents.

Can you test if that simpler code works in your context ?

Jo-NMS: Hi @g-bougard 

Thank you for the feedback.

It doesn’t work as suggested because the value lacks a space between the numeric part and the unit.

The getCanonicalSize() function regex requires a space between the number and the unit.

```suggestion
        if ($key eq 'lxc.cgroup.memory.limit_in_bytes' || $key eq 'lxc.cgroup2.memory.max') {
            if ($val =~ /[KMGTP]$/i) {
                $val .= " b";
            } elsif ($val =~ /^\d+$/) {
                $val .= " bytes";
            }
            $container->{MEMORY} = getCanonicalSize($val, 1024);
        }
```

This should work. Does it meet the requirements?

g-bougard: > The getCanonicalSize() function regex requires a space between the number and the unit.

I don't agree with that point. And even, l. 314 in that function spaces are removed.
Also in original code, you have `$val .= "b" if $val =~ /[KMGTP]$/i;` where now you add a space (removed later of course but so not coherent with previous code).

> It doesn’t work as suggested because the value lacks a space between the numeric part and the unit.

As just said, indeed, there's no such requirement in getCanonicalSize() so it should work. I tested myself to pass `16777216bytes` or `16777216 bytes` to the function and it always returns `16` as expected.

So I agree with that new version but don't add a space when a unit modifier still exists as this changes the previous behavior. I don't care for the space before `bytes` as this is a new case, this is up to you:
```suggestion
        if ($key eq 'lxc.cgroup.memory.limit_in_bytes' || $key eq 'lxc.cgroup2.memory.max') {
            if ($val =~ /[KMGTP]$/i) {
                $val .= "b";
            } elsif ($val =~ /^\d+$/) {
                $val .= " bytes";
            }
            $container->{MEMORY} = getCanonicalSize($val, 1024);
        }
```

Jo-NMS: After checking again I was wrong. It had nothing to do with the spacing, it was just a syntax issue on the regex missing the closing "/", so this is indeed better:

```suggestion
        if ($key eq 'lxc.cgroup.memory.limit_in_bytes' || $key eq 'lxc.cgroup2.memory.max') {
            $val .= $val =~ /[KMGTP]$/i ? "b" : $val =~ /^\d+$/ ? "bytes" : "";
            $container->{MEMORY} = getCanonicalSize($val, 1024);
        }
```

g-bougard: Oh yes, sorry for that stupid syntax error :innocent:

g-bougard:
Hi @Jo-NMS 

can you also fix tests ?

You should just have to fix 2 values in `t/tasks/inventory/virtualization/lxc.t`, on lines 26 & 58. On l.26, you have to replace `'2048000'` by `1`... yes, 2048000 bytes are rounded to 1Mb. And l.58, you can replace 2147483648 by 2048 which seems more coherent than the previous value.

PR #884, opened 2025-03-25
Update Hardware.pm to get UUID on global zones

topinet:
On modern Solaris servers, multiple LDOMs can be created linked to the same hardware, so the serial number is the same on all of them.

On my setup, GLPI got confused when updating inventory because multiple Solaris computers had the same serial number but none of them had a UUID.

GLPI agent remote inventory tries to get the UUID of the global zones using the command '/usr/sbin/zoneadm -z global list -p', but this returns an empty string when checked on the global zone (both on Solaris 10 and 11.4).

I'm not sure if this method would work on legacy Solaris, so I added an additional check for the global zone in sparc, to get a unique UUID from the command 'virtinfo -u' when '/usr/sbin/zoneadm -z global list -p' returns an empty string.

This was tested on Solaris 10 and 11.4.

### DIFF
--- a/lib/GLPI/Agent/Task/Inventory/Solaris/Hardware.pm
+++ b/lib/GLPI/Agent/Task/Inventory/Solaris/Hardware.pm
@@ -50,12 +50,30 @@ sub doInventory {
                 command => '/usr/sbin/zoneadm -z global list -p',
                 logger  => $logger
             );
+            if ($hardware->{UUID} eq '') {
+                $hardware->{UUID} = _getUUIDGlobal( logger  => $logger );
+            }
         }
     } elsif ($arch eq 'sparc') {
         $hardware->{UUID} = _getUUID( logger  => $logger );
     }
 
     $inventory->setHardware($hardware);
+}
+
+sub _getUUIDGlobal {
+    my (%params) = (
+        command => 'virtinfo -u',
+        @_
+    );
+
+    my $line = getFirstLine(%params);
+    return unless $line;
+
+    my @info = split(/: /, $line);
+    my $uuid = $info[1];
+
+    return $uuid;
 }
 
 sub _getUUID {
--- a/lib/GLPI/Agent/Task/Inventory/Solaris/Hardware.pm
+++ b/lib/GLPI/Agent/Task/Inventory/Solaris/Hardware.pm
@@ -50,7 +50,7 @@ sub doInventory {
                 command => '/usr/sbin/zoneadm -z global list -p',
                 logger  => $logger
             );
-            if ($hardware->{UUID} eq '') {
+            if (empty($hardware->{UUID})) {
                 $hardware->{UUID} = _getUUIDGlobal( logger  => $logger );
             }
         }
--- a/t/tasks/inventory/solaris/hardware.t
+++ b/t/tasks/inventory/solaris/hardware.t
@@ -11,9 +11,7 @@ use GLPI::Agent::Tools::Solaris;
 use GLPI::Agent::Task::Inventory::Solaris::Hardware;
 
 my %virtinfo_tests = (
-    'sample1' => {
-        'Domain UUID' => '915fbcf6-2b64-48ba-9b7b-05df341428be',
-    },
+    'sample1' => '915fbcf6-2b64-48ba-9b7b-05df341428be',
 );
 
 plan tests => 1;
--- a/t/tasks/inventory/solaris/hardware.t
+++ b/t/tasks/inventory/solaris/hardware.t
@@ -18,6 +18,6 @@ plan tests => (scalar keys %virtinfo_tests) + 1;
 
 foreach my $test (keys %virtinfo_tests) {
     my $file   = "resources/solaris/virtinfo/$test";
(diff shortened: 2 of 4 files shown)


lib/GLPI/Agent/Task/Inventory/Solaris/Hardware.pm
```diff
@@ -50,6 +50,9 @@ sub doInventory {
                 command => '/usr/sbin/zoneadm -z global list -p',
                 logger  => $logger
             );
+            if ($hardware->{UUID} eq '') {
+                $hardware->{UUID} = _getUUIDGlobal( logger  => $logger );
+            }
```
g-bougard: Here, testing with `eq`, you assume `$hardware->{UUID}` is always defined. But it is the result of `_getUUID()` function call. If for some reason `zoneadm` command doesn't return a line, `$hardware->{UUID}` will be not defined and your test will fail with a perl warning and that's not a clean processing and may not give the expected result.
For that kind of test, we have [`empty()` dedicated API in GLPI::Agent::Tools](https://github.com/glpi-project/glpi-agent/blob/a4b53380b557a19b7e90ccbb646734106294536f/lib/GLPI/Agent/Tools.pm#L154), you can use. So you should better use:
```suggestion
            if (empty($hardware->{UUID})) {
                $hardware->{UUID} = _getUUIDGlobal( logger  => $logger );
            }
```

g-bougard (changes requested):
Hi @topinet 

thank you for this nice PR.

Do you think you can provide some command output sample and create a test file ?
For the test file, `t/tasks/inventory/solaris/hardware.t` would be appropriate, and you can probably use `t/tasks/inventory/solaris/bios.t` as template.
Then you would need to add sample outputs as file under `resources/solaris` subfolders. For each of your cases, we need outputs for `/usr/sbin/zoneadm -z global list -p` and `/usr/sbin/zoneadm list -p` to be passed to `_getUUID()` in a `file` parameter and of course outputs of `virtinfo -u` to test results of `_getUUIDGlobal()`.

If you don't manage to generate the test file, I can do it if you add `resources` files. So I would only really request you to add that sample `resources` files.

You can of course change sensible datas in `resources` file. But in that case, at least keep the same format, this important to validate parsing.

topinet:
Hi @g-bougard, please review carefully so it's my first perl test file.
Regards

t/tasks/inventory/solaris/hardware.t
```diff
@@ -0,0 +1,25 @@
...
+use Test::NoWarnings;
+
+use GLPI::Agent::Tools::Solaris;
+use GLPI::Agent::Task::Inventory::Solaris::Hardware;
+
+my %virtinfo_tests = (
+    'sample1' => {
+        'Domain UUID' => '915fbcf6-2b64-48ba-9b7b-05df341428be',
+    },
+);
```
g-bougard: As `_getUUIDGlobal()` returns just UUID, you should simplify:
```suggestion
my %virtinfo_tests = (
    'sample1' => '915fbcf6-2b64-48ba-9b7b-05df341428be',
);
```

t/tasks/inventory/solaris/hardware.t
```diff
@@ -0,0 +1,25 @@
...
+
+plan tests => (scalar keys %virtinfo_tests) + 1;
+
+foreach my $test (keys %virtinfo_tests) {
+    my $file   = "resources/solaris/virtinfo/$test";
+    my $result = GLPI::Agent::Task::Inventory::Solaris::Hardware::_getUUIDGlobal(file => $file);
+    cmp_deeply($result, $virtinfo_tests{$test}, "virtinfo parsing: $test");
+}
```
g-bougard: Here as you only have to test a scalar result, `is()` should better be used:
```suggestion
foreach my $test (keys %virtinfo_tests) {
    my $file   = "resources/solaris/virtinfo/$test";
    my $result = GLPI::Agent::Task::Inventory::Solaris::Hardware::_getUUIDGlobal(file => $file);
    is($result, $virtinfo_tests{$test}, "virtinfo parsing: $test");
}
```

g-bougard (changes requested):
Hi @topinet 

You are near to have a good test.

As I still said, I would also require zoneadm outputs in the same context.

A last point, it would also be better to use another test name than "sample1". A name giving few info about the context, for example Solaris version would be better.

t/tasks/inventory/solaris/hardware.t
```diff
@@ -0,0 +1,23 @@
...
+my %virtinfo_tests = (
+    'solaris' => '915fbcf6-2b64-48ba-9b7b-05df341428be',
+);
+
+plan tests => (scalar keys %virtinfo_tests) + 1;
+
+foreach my $test (keys %virtinfo_tests) {
+    my $file   = "resources/solaris/virtinfo/$test";
+    my $result = GLPI::Agent::Task::Inventory::Solaris::Hardware::_getUUIDGlobal(file => $file);virtinfo_tests
+    is($result, $virtinfo_tests{$test}, "virtinfo parsing: $test");
```
g-bougard: There's a `virtinfo_tests` at the end of line 21. Probably a mis-tape, just remove it.

PR #944, opened 2025-05-26
Fix error when DB name is only a number during MSSQL database inventory

SL-Gundam:
When the database name is only a number, which is often the case with Exact Globe (Next)+, The script will produce errors when retrieving information about the database.
This fixes those errors

### DIFF
--- a/lib/GLPI/Agent/Task/Inventory/Generic/Databases/MSSQL.pm
+++ b/lib/GLPI/Agent/Task/Inventory/Generic/Databases/MSSQL.pm
@@ -126,7 +126,7 @@ sub _getDatabaseService {
                 or next;
 
             my ($size) = _runSql(
-                sql => "USE $db_name ; EXEC sp_spaceused",
+                sql => "USE [$db_name] ; EXEC sp_spaceused",
                 %params
             ) =~ /^$db_name;([0-9.]+\s*\S+);/;
             if ($size) {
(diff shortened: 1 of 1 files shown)


g-bougard:
Hi @SL-Gundam 

thank you for your PR. I understand the problem, but I'm not a MSSQL expert. Can you point me to an official documentation page explaining this syntax ?

SL-Gundam:
In **SQL Server (MSSQL)**, you use **square brackets** around a **database name** in the `USE` command only when the name contains:

1. **Spaces**
2. **Special characters** (e.g., hyphens, periods, etc.)
3. **Reserved SQL keywords**

### ✅ Examples (when brackets are needed):

```sql
USE [My Database];         -- contains a space
USE [Sales-2023];          -- contains a hyphen
USE [Order];               -- "Order" is a reserved keyword
```

### ✅ Examples (when brackets are optional):

```sql
USE Sales;                 -- simple, no spaces or special characters
USE InventoryDB;           -- safe, no conflicts
```

### 🧠 Why use brackets?

Brackets tell SQL Server to treat the enclosed value as an **identifier**, even if it would otherwise be invalid or ambiguous.

### 📝 General rule:

If you're ever unsure, using brackets is always safe. But they're only strictly necessary when the database name isn't a valid regular identifier.

Source: https://learn.microsoft.com/en-us/sql/relational-databases/databases/database-identifiers?view=sql-server-ver17
This describes it more for column names but the principle is the same.

g-bougard (approved):
Thank you for the documentation link.

Issue #1009, opened 2025-09-11
qemu Virtual Machine Inventory doesn't work


Megachip:
### Bug reporting acknowledgment

Yes, I read it

### Professional support

Yes, I know

### Describe the bug

Running an remote Inventory (as root) on an Ubuntu but didn't see running qemu Virtual Maschines (except the docker ones)

### To reproduce

1. running remote inventory task

### Expected behavior

See all running VMs, not only Docker container

### Operating system

Linux

### GLPI Agent version

v1.15

### GLPI version

10.0.18

### GLPIInventory plugin or other plugin version

GLPI Inventory v1.5.3

### Additional context

```
[debug] Running GLPI::Agent::Task::Inventory::Virtualization
[debug] Running GLPI::Agent::Task::Inventory::Virtualization::Docker
```
Seems qemu isn't even detected on the system.
Its an OpenStack Compute Node

```
ps -ef | grep qemu
[...]

PR #1010, opened 2025-09-12
Adding qemu CPU number detection

Megachip:


lib/GLPI/Agent/Task/Inventory/Virtualization/Qemu.pm
```diff
@@ -47,6 +47,11 @@ sub _parseProcessList {
             $values->{uuid} = $1;
         } elsif ($option =~ m/^enable-kvm/) {
```
Megachip, line 48: Maybe this should also be reworked/checked: `accel=kvm`

lib/GLPI/Agent/Task/Inventory/Virtualization/Qemu.pm
```diff
@@ -94,7 +99,7 @@ sub doInventory {
             entry => {
                 NAME      => $values->{name},
                 UUID      => $values->{uuid},
-                VCPU      => 1,
+                VCPU      => $values->{vcpu},
```
g-bougard: You can't trust `vcpu` is always defined. You should set 1 as default if not:
```suggestion
                VCPU      => $values->{vcpu} || 1,
```

g-bougard:
Have you some output samples to test with ?

lib/GLPI/Agent/Task/Inventory/Virtualization/Qemu.pm
Megachip: Shouldn't we set this to 0 if there is no CPU option? Like in docker? On the other hand, I'll guess 1 is default ;)

Megachip: Shouldn't the default set in the same way like on memory?

Megachip:
> Have you some output samples to test with ?

Tested it with about 100 ... atm you can use the one in issue #1009

lib/GLPI/Agent/Task/Inventory/Virtualization/Qemu.pm
```diff
@@ -47,6 +47,11 @@ sub _parseProcessList {
             $values->{uuid} = $1;
         } elsif ($option =~ m/^enable-kvm/) {
             $values->{vmtype} = "kvm";
+        } elsif ($option =~ m/^smp ([^\s,]+)/) {
+            my ($cpus) = split(/,/, $1);
+            if ($cpus =~ /^\d+$/) {
+                $values->{vcpu} = int($cpus);
+            }
```
g-bougard: Reading again this condition, I'm perplex. To me the regexp on `$option` means you take everything not a space or a comma after `smp` option. And after you're splitting against comma (which shouldn't be there) so. I found also in qemu help that the number can be set with: `cpus=2` for 2 cpus as example. Then I think a better code here would be:
```suggestion
        } elsif ($option =~ m/^smp (\S+)$/) {
            my ($cpus) = grep { /^(?:cpus=)?\d+$/ } split(/,/, $1);
            if ($cpus && $cpus =~ /(\d+)$/) {
                $values->{vcpu} = int($1);
            }
```
This code will also support to set `cpus=2` not at first place.

Megachip: Sounds reasonable. But sould we think a bit wider? what about:
`-smp cores=2,threads=2,sockets=1` How did GLPI define CPU?

Megachip:
Maybe we should use something like `core*sockets*threads`?

g-bougard:
Hi @Megachip 

please open another discussion and be more accurate than just pushing your thought without any context.

PR #1012, opened 2025-09-15
Add --category option

0xff8:
Here is an initial proposal for a new `--category` option, that if present - will enable only desired metrics and exclude everything else.

As continuation of https://github.com/glpi-project/glpi-agent/discussions/1002

### DIFF
--- a/lib/GLPI/Agent/Task/Inventory.pm
+++ b/lib/GLPI/Agent/Task/Inventory.pm
@@ -159,6 +159,20 @@ sub run {
         map { $_ => 1 } @{$self->{config}->{'category'}}
     };
 
+    if (keys(%{$self->{enabled}})) {
+        # Be sure at least bios and hardware category are enabled via category option
+        foreach my $category (qw(bios hardware)) {
+            $self->{logger}->debug("$category not enabled, forcing as it is required")
+                unless $self->{enabled}->{$category};
+            $self->{enabled}->{$category} = 1;
+        }
+        # Software category requires os to be enabled too
+        if ($self->{enabled}->{software} && !$self->{enabled}->{os}) {
+            $self->{logger}->debug("Forcing os category as required by software one")
+            $self->{enabled}->{os} = 1;
+        }
+    }
+
     # Support inventory event
     if ($event && !$self->setupEvent()) {
         $self->{logger}->info("Skipping Inventory task event on ".$self->{target}->id());
--- a/lib/GLPI/Agent/Task/Inventory.pm
+++ b/lib/GLPI/Agent/Task/Inventory.pm
@@ -37,7 +37,7 @@ sub isEnabled {
                 my %disabled = map { $_ => 1 } @{$self->{config}->{'no-category'}};
                 my %enabled = map { $_ => 1 } @{$self->{config}->{'category'}};
                 my @params;
-            my $cant_load_glpi_client = 0;
+                my $cant_load_glpi_client = 0;
                 foreach my $param (@{$tasks->{inventory}->{params}}) {
                     my @validated;
                     if (!$param->{category} || $disabled{$param->{category}} || (%enabled && ! exists $enabled{ $param->{category} })) {
--- a/lib/GLPI/Agent/Task/Inventory.pm
+++ b/lib/GLPI/Agent/Task/Inventory.pm
@@ -35,11 +35,12 @@ sub isEnabled {
                 # Add a GLPI client to each param with a category and a use property
                 # and if related category is not disabled
                 my %disabled = map { $_ => 1 } @{$self->{config}->{'no-category'}};
+                my %enabled = map { $_ => 1 } @{$self->{config}->{'category'}};
                 my @params;
-                my $cant_load_glpi_client = 0;
+            my $cant_load_glpi_client = 0;
                 foreach my $param (@{$tasks->{inventory}->{params}}) {
                     my @validated;
-                    if (!$param->{category} || $disabled{$param->{category}}) {
+                    if (!$param->{category} || $disabled{$param->{category}} || (%enabled && ! exists $enabled{ $param->{category} })) {
                     } elsif ($param->{params_id}) {
                         # Here we must handle the case of remotely triggered events
                         my @categories = map { trimWhitespace($_) } split(/,+/, $param->{category});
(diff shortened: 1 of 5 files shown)


0xff8:
Hi @g-bougard ,

do you want deeper integration (installers + configs + per module check) too ?

lib/GLPI/Agent/Task/Inventory.pm
```diff
@@ -413,7 +417,7 @@ sub _initModulesList {
         if (defined(*{$module."::category"})) {
             no strict 'refs'; ## no critic (ProhibitNoStrict)
             my $category = &{$module."::category"}();
-            if ($category && $self->{disabled}->{$category}) {
+            if ($category && ($self->{disabled}->{$category} || ( %{ $self->{enabled} } && !exists $self->{enabled}->{$category} ))) {
```
g-bougard, line 434: Here I would prefer a `elsif` after the original case with a log telling category is "not enabled" in place of "disabled". This way, we can better understand why a module remains disabled while reading the log.

lib/GLPI/Agent/Task/Inventory.pm
```diff
@@ -532,6 +536,7 @@ sub _runModule {
             datadir       => $self->{datadir},
             inventory     => $self->{inventory},
             no_category   => $self->{disabled},
+            category      => $self->{enabled},
```
g-bougard, line 553: Not needed, no module uses `category` param in their `isEnabled` method.

g-bougard (changes requested):
Hi @0xff8,

I'm still perplex.

I see few problem with this option. It can break inventory if miss-used.

If I merge it, I'll modify few things to force required category. For example, bios and hardware categories are required for glpi and os is required if software category is enabled (see https://github.com/glpi-project/glpi-agent/blob/75a8699a5ab0d4a489e50a66bad6a4e4948dcca5/lib/GLPI/Agent/Task/Inventory.pm#L235).

A problem is still this option could be very long as the default would be: `accesslog,antivirus,battery,bios,controller,cpu,database,drive,environment,firewall,hardware,input,licenseinfo,local_group,local_user,lvm,memory,modem,monitor,network,os,port,printer,process,provider,psu,registry,remote_mgmt,rudder,slot,software,sound,storage,usb,user,video,virtualmachine`

Maybe another way would be to use a `categories` option which would support some keywords with preselected category values: `full`, `system`, `basic`, `extended` or what ever people can think of.

0xff8:
thank you very much for the review @g-bougard ! 

My initial goal was not to change the existing behavior. So, if no `--category` parameter is used the `glpi-agent` behaves as usual. I now see that some fields are required for partial updates, and as a simplest fix, I can declare the new `--category` option to be used only with the `--full` option - so partial updates won't be affected at all. Or I can extend current code (and make logic more difficult) and track dependencies in modules for partial updates.

> A problem is still this option could be very long as the default would be: accesslog,antivirus,battery,bios,controller,cpu,database,drive,environment,firewall,hardware,input,licenseinfo,local_group,local_user,lvm,memory,modem,monitor,network,os,port,printer,process,provider,psu,registry,remote_mgmt,rudder,slot,software,sound,storage,usb,user,video,virtualmachine

By default, the plan is to request only the privacy-friendly data, i.e. `bios`,`cpu`,`ram`,`battery`, `storage` . I see this option as the very precise filter to get as minimum information as necessary. Also, I wanted this option to be compatible with `--no-category` option , because users know it and its format.

As an example, currently, the command we use looks like: `glpi-agent --no-category process,environment,local_user,local_group,accesslog,network,software,video,battery,psu,monitor,user,firewall--server https://internal.server.local --full` . it is quite long already :)

I am happy if you share your vision about `--category` option behavior - my understanding is - it will make partial updates unnecessary, since we already know that we request a very minimal set of the metrics.   
Anyway, I am ready to adjust the implementation accordingly to get it merged 🚀

lib/GLPI/Agent/Task/Inventory.pm
```diff
@@ -154,6 +155,9 @@ sub run {
     $self->{disabled}  = {
         map { $_ => 1 } @{$self->{config}->{'no-category'}}
     };
+    $self->{enabled}  = {
+        map { $_ => 1 } @{$self->{config}->{'category'}}
+    };
```
g-bougard, line 160: One point, some fields are required so GLPI can import the inventory, this is not specific to partial inventory. Actually, if you set `no-category` to `bios,hardware`, inventory won't be imported.

You can probably set here some simple constraint like this:
```suggestion
    $self->{enabled}  = {
        map { $_ => 1 } @{$self->{config}->{'category'}}
    };

    if (keys(%{$self->{enabled}})) {
        # Be sure at least bios and hardware category are enabled via category option
        foreach my $category (qw(bios hardware)) {
            $self->{logger}->debug("$category not enabled, forcing as it is required")
                unless $self->{enabled}->{$category};
            $self->{enabled}->{$category} = 1;
        }
        # Software category requires os to be enabled too
        if ($self->{enabled}->{software} && !$self->{enabled}->{os}) {
            $self->{logger}->debug("Forcing os category as required by software one")
            $self->{enabled}->{os} = 1;
        }
    }
```
I think this is sufficient, can you test it ?

0xff8:
I haven't finished yet, sorry, I found after merging your changes, not all reports are working properly. Please, let me spent a bit more time on this .

lib/GLPI/Agent/Task/Inventory.pm
```diff
@@ -154,6 +155,23 @@ sub run {
...
+    if (keys(%{$self->{enabled}})) {
+        # Be sure at least bios and hardware category are enabled via category option
+        foreach my $category (qw(bios hardware)) {
+            $self->{logger}->debug("$category not enabled, forcing as it is required")
+                unless $self->{enabled}->{$category};
+            $self->{enabled}->{$category} = 1;
+        }
+        # Software category requires os to be enabled too
+        if ($self->{enabled}->{software} && !$self->{enabled}->{os}) {
+            $self->{logger}->debug("Forcing os category as required by software one")
```
g-bougard, line 171: My apologies, one char is missing at the end here:
```suggestion
            $self->{logger}->debug("Forcing os category as required by software one");
```

PR #1051, opened 2025-11-19
Add support for ssl_key_file (as required for client certificate auth)

jmguzmanc:
This patch allows the client to pass the SSL key file (along with the already provided SSL certificate file) to the LWP client, so that we can use a custom SSL certificate, as required by the server when client certificate authentication is required.

### DIFF
--- a/lib/GLPI/Agent/HTTP/Client.pm
+++ b/lib/GLPI/Agent/HTTP/Client.pm
@@ -44,7 +44,7 @@ sub new {
         if $ssl_cert_file && ! -f $ssl_cert_file;
 
     my $ssl_key_file = $params{ssl_key_file} || $config->{'ssl-key-file'};
-    die "non-existing client certificate file $ssl_key_file"
+    die "non-existing client private key file $ssl_key_file"
         if $ssl_key_file && ! -f $ssl_key_file;
 
     # We should still keep SSL certs cache if running in long running netdiscovery
--- a/lib/GLPI/Agent/HTTP/Client.pm
+++ b/lib/GLPI/Agent/HTTP/Client.pm
@@ -43,6 +43,10 @@ sub new {
     die "non-existing client certificate file $ssl_cert_file"
         if $ssl_cert_file && ! -f $ssl_cert_file;
 
+    my $ssl_key_file = $params{ssl_key_file} || $config->{'ssl-key-file'};
+    die "non-existing client certificate file $ssl_key_file"
+        if $ssl_key_file && ! -f $ssl_key_file;
+
     # We should still keep SSL certs cache if running in long running netdiscovery
     # or netinventory task with expiration set in a dedicated thread
     $_SSL_ca->{_expiration} = getExpirationTime()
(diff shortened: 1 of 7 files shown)


g-bougard:
Hi @jmguzmanc

have you tried to merge your key file in PEM format into the file set on SSL_cert_file ?
If I remember well, IO::Socket::SSL will try to read key file from it as final chance to get it when ssl_key_file is not set.

Anyway, I agree having the option should be a good thing as this could permit to set dedicated rights on related file.

lib/GLPI/Agent/HTTP/Client.pm
```diff
@@ -43,6 +43,10 @@ sub new {
     die "non-existing client certificate file $ssl_cert_file"
         if $ssl_cert_file && ! -f $ssl_cert_file;
 
+    my $ssl_key_file = $params{ssl_key_file} || $config->{'ssl-key-file'};
+    die "non-existing client certificate file $ssl_key_file"
```
g-bougard: I see there a little mistake: the text is just not correct as a copy/paste of the previous test even if it doesn't include the same file value:
```suggestion
    die "non-existing client private key file $ssl_key_file"
```

g-bougard (changes requested):
Hi @jmguzmanc 

thank you for your suggestion.

Can you check my suggestion on the little mistake for the "die" missing file condition ?

Can you also update concerned scripts under `bin` folder so the configuration could also be used as option ? Don't worry anyway, I'll do it myself if you don't feel easy enough with that changes.

lib/GLPI/Agent/HTTP/Client.pm
jmguzmanc: Nice catch! fixing

jmguzmanc:
@g-bougard, I confirm that combining private and cert in the same file, works... but it was impossible to me to know it.

So, keeping the suggestion fro this patch, I added some extra documentation, and the options in the script.

lib/GLPI/Agent/HTTP/Protocol/https.pm
```diff
@@ -16,6 +16,8 @@ sub import {
         if $params{ca_cert_dir};
     IO::Socket::SSL::set_ctx_defaults(ssl_cert_file => $params{ssl_cert_file})
         if $params{ssl_cert_file};
+    IO::Socket::SSL::set_ctx_defaults(ssl_key_file => $params{ssl_key_file})
+        if $params{ssl_key_file};
```
g-bougard, line 20: Here, I'm okay, but you also have to set ssl_key_file in the caller which for glpî-agent happens from lib/GLPI/Agent/HTTP/Client.pm around line 541. You should there insert a line at l.545 with:
```perl
                ssl_key_file => $self->{ssl_key_file},
```

This case is only for older environments... so no body should be concerned, but who knows ? ;-)

jmguzmanc: Absolutelly right.. I will do.. Apologies, I am not familiar with the whole source code yet. Thanks for catching this.

g-bougard (approved):
Hi @jmguzmanc 

thank you for your work.

I'll update myself dedicated support in windows installer and documentation after the merge.

PR #1076, opened 2026-01-02
Enhance VmWareDesktop.pm to support Workstation

Megachip:
Generic Windows Support plus setting `VCPU` correcly.
Solving  #1075

### DIFF
--- a/lib/GLPI/Agent/Task/Inventory/Virtualization/VmWareDesktop.pm
+++ b/lib/GLPI/Agent/Task/Inventory/Virtualization/VmWareDesktop.pm
@@ -59,7 +59,7 @@ sub _getMachines {
     # skip first line
     shift @lines;
     
-    my $subsystem = ($^O eq 'MSWin32') ? "VmWare Workstation" : "VmWare Fusion";
+    my $subsystem = (OSNAME eq 'MSWin32') ? "VmWare Workstation" : "VmWare Fusion";
     my @machines;
     foreach my $line (@lines) {
         next unless has_file($line);
--- a/lib/GLPI/Agent/Task/Inventory/Virtualization/VmWareDesktop.pm
+++ b/lib/GLPI/Agent/Task/Inventory/Virtualization/VmWareDesktop.pm
@@ -68,7 +68,7 @@ sub _getMachines {
 
         my $machine = {
             NAME      => $info{'displayName'} // $info{'displayname'},
-            VCPU      => $info{'numvcpus'},
+            VCPU      => $info{'numvcpus'} // 1,
             UUID      => $info{'uuid.bios'},
             MEMORY    => $info{'memsize'},
             STATUS    => STATUS_RUNNING,
--- a/lib/GLPI/Agent/Task/Inventory/Virtualization/VmWareDesktop.pm
+++ b/lib/GLPI/Agent/Task/Inventory/Virtualization/VmWareDesktop.pm
@@ -73,7 +73,7 @@ sub _getMachines {
             STATUS    => STATUS_RUNNING,
             SUBSYSTEM => $subsystem,
             VMTYPE    => "VmWare",
-            COMMENT   => $info{'annotation'},
+            COMMENT   => $info{'annotation'} // "",
         };
 
         push @machines, $machine;
--- a/lib/GLPI/Agent/Task/Inventory/Virtualization/VmWareDesktop.pm
+++ b/lib/GLPI/Agent/Task/Inventory/Virtualization/VmWareDesktop.pm
@@ -67,7 +67,7 @@ sub _getMachines {
         my %info = _getMachineInfo(file => $line, logger => $params{logger});
 
         my $machine = {
-            NAME      => $info{'displayname'},
+            NAME      => $info{'displayName'} // $info{'displayname'},
             VCPU      => $info{'numvcpus'},
             UUID      => $info{'uuid.bios'},
             MEMORY    => $info{'memsize'},
--- a/lib/GLPI/Agent/Task/Inventory/Virtualization/VmWareDesktop.pm
+++ b/lib/GLPI/Agent/Task/Inventory/Virtualization/VmWareDesktop.pm
@@ -33,7 +33,6 @@ sub doInventory {
         $command = "'/Library/Application Support/VMware Fusion/vmrun' list";
     }
     elsif (canRun('C:\\\\Program Files (x86)\\\\VMware\\\\VMware Workstation\\\\vmrun.exe')) {
-        # specify Workstation target with -T ws
         $command = '"C:\\\\Program Files (x86)\\\\VMware\\\\VMware Workstation\\\\vmrun.exe" list';
     }
     else {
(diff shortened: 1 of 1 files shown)


lib/GLPI/Agent/Task/Inventory/Virtualization/VmWareDesktop.pm
```diff
@@ -43,20 +58,21 @@ sub _getMachines {
 
     # skip first line
     shift @lines;
-
+    
+    my $subsystem = ($^O eq 'MSWin32') ? "VmWare Workstation" : "VmWare Fusion";
```
g-bougard: To make it working via RemoteInventory task, use `OSNAME` api provided by GLPI::Agent::Tools:
```suggestion
    my $subsystem = (OSNAME eq 'MSWin32') ? "VmWare Workstation" : "VmWare Fusion";
```

lib/GLPI/Agent/Task/Inventory/Virtualization/VmWareDesktop.pm
```diff
@@ -43,20 +58,21 @@ sub _getMachines {
...
     my @machines;
     foreach my $line (@lines) {
         next unless has_file($line);
 
         my %info = _getMachineInfo(file => $line, logger => $params{logger});
 
         my $machine = {
-            NAME      => $info{'displayName'},
-            VCPU      => 1,
+            NAME      => $info{'displayname'},
```
g-bougard: I suspect here the original field may exist in other OS like OSX. Do you think you can have `vmrun list` output for OSX and Windows to compare ?

To not break anything, the best is to use the first field and fallback on the one you think we must use with the following code:
```suggestion
            NAME      => $info{'displayName'} // $info{'displayname'},
```
It means "use the other field if the first is not defined".

lib/GLPI/Agent/Task/Inventory/Virtualization/VmWareDesktop.pm
```diff
@@ -43,20 +58,21 @@ sub _getMachines {
...
+            VCPU      => $info{'numvcpus'},
```
g-bougard: For the VCPU case, if the field is not defined, we have to fallbacl on the original value:
```suggestion
            VCPU      => $info{'numvcpus'} // 1,
```

lib/GLPI/Agent/Task/Inventory/Virtualization/VmWareDesktop.pm
```diff
@@ -14,7 +14,8 @@ use GLPI::Agent::Tools::Virtualization;
 sub isEnabled {
     return
         canRun('/Library/Application Support/VMware Fusion/vmrun') ||
-        canRun('vmrun');
+        canRun('vmrun') ||
+        canRun('C:\\\\Program Files (x86)\\\\VMware\\\\VMware Workstation\\\\vmrun.exe');
```
g-bougard: I'm curious: does VMware only provide 32-bits version installation ? If a 64-bits version exists, you have to handle it too.

Can you also try by using "/" path separator ? This should be supported by perl, even on windows.

Eventually to simplify code, you may want to use constant this way:
```perl
...

use constant WINDOW_VMRUN => 'C:/Program Files (x86)/VMware/VMWare Workstation/vmrun.exe';

sub isEnabled {
...
        canRun(WINDOWS_VMRUN);
...
        $command = '"'.WINDOWS_VMRUN.'" list';
...
```

Also, can you tell if the installation path can be changed ? In that case, we can't use a static file path.

g-bougard (changes requested):
Hi @Megachip 

I see few little problems with few required changes. Can you check my comments ?

Thank you

lib/GLPI/Agent/Task/Inventory/Virtualization/VmWareDesktop.pm
Megachip: Finally we should find a better solution, cause beside fusion and Worksstation (ws) theres also player als subsystem.

lib/GLPI/Agent/Task/Inventory/Virtualization/VmWareDesktop.pm
Megachip: Good question. But I would expect that VMX files should be same on all Platforms. Didn't find any example with upper case.

lib/GLPI/Agent/Task/Inventory/Virtualization/VmWareDesktop.pm
Megachip: whyever they are installed in (x86) even on 64 bit systems. Will check custom installation pathes.

lib/GLPI/Agent/Task/Inventory/Virtualization/VmWareDesktop.pm
```diff
@@ -43,21 +57,23 @@ sub _getMachines {
...
-            VCPU      => 1,
+            NAME      => $info{'displayName'} // $info{'displayname'},
+            VCPU      => $info{'numvcpus'} // 1,
             UUID      => $info{'uuid.bios'},
             MEMORY    => $info{'memsize'},
             STATUS    => STATUS_RUNNING,
-            SUBSYSTEM => "VmWare Fusion",
+            SUBSYSTEM => $subsystem,
             VMTYPE    => "VmWare",
+            COMMENT   => $info{'annotation'},
```
g-bougard: For the added "COMMENT" field, you must handle the case it is not defined to avoid perl warning when not defined:
```suggestion
            COMMENT   => $info{'annotation'} // "",
```

lib/GLPI/Agent/Task/Inventory/Virtualization/VmWareDesktop.pm
Megachip: [debug] module GLPI::Agent::Task::Inventory::Virtualization::VmWareDesktop disabled: failure to load (Bareword "WINDOWS_VMRUN" not allowed while "strict subs" in use at C:/Apps/GAP/nightly/perl/agent/GLPI/Agent/Task/Inventory/Virtualization/VmWareDesktop.pm line 18.

lib/GLPI/Agent/Task/Inventory/Virtualization/VmWareDesktop.pm
```diff
@@ -23,8 +24,21 @@ sub doInventory {
     my $inventory = $params{inventory};
     my $logger    = $params{logger};
 
-    my $command = canRun('vmrun') ?
-        'vmrun list' : "'/Library/Application Support/VMware Fusion/vmrun' list";
+    my $command;
+
```
g-bougard, line 28: Please remove white spaces here.
```suggestion

```

lib/GLPI/Agent/Task/Inventory/Virtualization/VmWareDesktop.pm
```diff
@@ -43,21 +57,23 @@ sub _getMachines {
 
     # skip first line
     shift @lines;
-
+
```
g-bougard, line 60: Please remove white spaces here.
```suggestion

```

g-bougard (changes requested):
Unittests are not happy with white spaces you leaved for indentation.

PR #1118, opened 2026-03-04
Refactor Fortinet module to support newer FortiAP series using is() API

tmayans:


lib/GLPI/Agent/SNMP/MibSupport/Fortinet.pm
```diff
@@ -22,10 +22,17 @@ use constant fgHaStatsIndex     => fgHaStatsEntry . '.1';
...
+use constant fnApGFirmware  => fnFortiAPMib . '.1.1.0';
+
 our $mibSupport = [
     {
         name        => "fortinet",
-        sysobjectid => getRegexpOidMatch(fnFortiGateMib)
+	# Added .120 branch for FortiAP (F and G series)
+	match       => qr/^12356\.(101|120)\./,
+        # sysobjectid => getRegexpOidMatch(fnFortiGateMib . "|" . fnFortiAPMib)
     }
```
g-bougard: Here, you shouldn't merge the test in a regexp but separate it to have another "mibsupport" case:
```suggestion
        sysobjectid => getRegexpOidMatch(fnFortiGateMib)
    },
    # Added .120 branch for FortiAP (F and G series)
    {
        name        => "fortiAP",
        sysobjectid => getRegexpOidMatch(fnFortiAPMib)
    }
```

lib/GLPI/Agent/SNMP/MibSupport/Fortinet.pm
```diff
@@ -66,9 +73,24 @@ sub getComponents {
 sub getSerial {
     my ($self) = @_;
 
+    # (FortiAP F/G)
+    if ($self->is(fnFortiAPMib)) {
```
g-bougard: The `is()` method takes mibsupport case as argument:
```suggestion
    if ($self->is("fortiAP")) {
```

lib/GLPI/Agent/SNMP/MibSupport/Fortinet.pm
```diff
@@ -66,9 +73,24 @@ sub getComponents {
...
     return getCanonicalString($self->get(fnSysSerial));
 }
 
+sub getFirmware {
+    my ($self) = @_;
+
+    # Si és FortiAP branca .120
+    if ($self->is(fnFortiAPMib)) {
+        return getCanonicalString($self->get(fnApGFirmware));
+    }
```
g-bougard: The `is()` method takes mibsupport case as argument and you shouldn't use another language than english in comments. Also as this is the only case in getFirmware, this should also be the last case that can return a defined value. Otherwise, it will return `0` for other devices which is the result of the test:
```suggestion
    return unless $self->is("fortiAP");

    return getCanonicalString($self->get(fnApGFirmware));
```

g-bougard (changes requested):
Hello @tmayans 

I see few little problem and misunderstanding in your code.
Plese can you follow my suggestions and check this still works on your devices.

g-bougard:
I noticed you also used tabulations for code indentation. This is not a big deal regarding perl execution, but on this project, we only use spaces. Can you take care about that point ? Thank you ;-)

lib/GLPI/Agent/SNMP/MibSupport/Fortinet.pm
```diff
@@ -66,9 +76,23 @@ sub getComponents {
 sub getSerial {
     my ($self) = @_;
 
+    # (FortiAP F/G)
+    if ($self->is("fortiAP")) {
+        return getCanonicalString($self->get(fnApGSerial));
+    }
```
g-bougard, line 85: After more thinking, we may need here to fallback on fnSysSerial oid which still may be supported. Here is a safer version:
```suggestion
    if ($self->is("fortiAP")) {
        my $serial = getCanonicalString($self->get(fnApGSerial));
        return $serial unless empty($serial);
    }
```

lib/GLPI/Agent/SNMP/MibSupport/Fortinet.pm
```diff
@@ -66,9 +78,26 @@ sub getComponents {
 sub getSerial {
     my ($self) = @_;
 
+    if ($self->is("fortiAP")) {
+	    my $serial = getCanonicalString($self->get(fnApGSerial));
+            return getCanonicalString($serial) if $serial;
```
g-bougard, line 83: Don't keep this line: calling again getCanonicalString() is a mistake and the next line does exactly the same.
```suggestion
```

lib/GLPI/Agent/SNMP/MibSupport/Fortinet.pm
```diff
@@ -66,9 +78,26 @@ sub getComponents {
...
+    }
     return getCanonicalString($self->get(fnSysSerial));
 }
 
+sub getFirmware {
+    my ($self) = @_;
+
+    if ($self->is(fnFortiAPMib)) {
+        return getCanonicalString($self->get(fnApGFirmware));
+    }
```
g-bougard, line 94: I think you made some errors while rebasing and forcing a push. These lines are wrong and were still replaced by the following ones.

g-bougard:
Hi @tmayans 

without your feedback, I merged manually your PR fixing mistakes I detected.

Please make test with nightly build and open an issue if you don't obtain the expected result.

PR #1163, opened 2026-05-18
feat(hyperv): collect DRIVES, SERIAL, MAC, IPADDRESS and OPERATINGSYSTEM for Hyper-V VMs

javierlago-ticgal:
## Summary

Enhances the Hyper-V inventory module to collect extended fields for each
virtual machine:

- **DRIVES**: list of attached virtual hard disks with path and provisioned size in MB
- **IPADDRESS**: guest IPv4 address via Hyper-V Integration Services (KVP)
- **OPERATINGSYSTEM**: guest OS name and version via KVP
- **SERIAL**: VM BIOS serial number
- **MAC**: first network adapter MAC address

## Motivation

When GLPI receives a Hyper-V VM inventory, it has no information about the
virtual disks attached to each VM, its network identity or its guest OS.
This change makes that data available without requiring any modification to
the GLPI server side.

Additionally, customer GLPI rules match VMs by UUID + serial number to detect
existing entries. Without `SERIAL` (and `MAC` as secondary identifier), the
rule can fail to match and generate duplicate VM entries in the inventory.

### DRIVES

Queries `MSVM_StorageAllocationSettingData` (ResourceType 31) on both WMI
namespaces (`root/virtualization/v2` and the legacy `root/virtualization`).
Each disk entry includes:

| Field    | Source                        | Description               |
|----------|-------------------------------|---------------------------|
| `VOLUMN` | `HostResource`                | Full path to the VHD/VHDX |
| `TOTAL`  | PowerShell `Get-VHD` → `Size` | Provisioned size in MB    |

Size is retrieved via `Get-VHD` on the host (requires Hyper-V PowerShell
module). If `Get-VHD` is unavailable, `TOTAL` defaults to `0`.

Path quoting uses PowerShell single-quoted strings to prevent `$` expansion
in UNC admin-share paths (e.g. `\\server\C$\VMs\vm.vhdx`).

Requires GLPI ≥ 10.0.25.

### IPADDRESS and OPERATINGSYSTEM

Queries `Msvm_KvpExchangeComponent` (Hyper-V Integration Services guest data).
Falls back to `Msvm_GuestNetworkAdapterConfiguration` for hosts where the
`NetworkAddressIPv4` KVP key is deprecated.

Requires GLPI ≥ 10.0.25.

### SERIAL

Collected from `MSVM_VirtualSystemSettingData.BIOSSerialNumber`, extending
the existing query already done for `BIOSGUID`. No GLPI version gate required.

### MAC

First MAC address from `Msvm_SyntheticEthernetPortSettingData`, formatted as
`XX:XX:XX:XX:XX:XX`. Falls back to `Msvm_EmulatedEthernetPortSettingData` for
Generation 1 VMs that use legacy emulated adapters. No GLPI version gate required.

## Changes

- `lib/GLPI/Agent/Task/Inventory/Virtualization/HyperV.pm`: new WMI queries
  and field assignments
- `lib/GLPI/Agent/Inventory.pm`: `DRIVES` declared as allowed field in
  `VIRTUALMACHINES`
- `t/tasks/inventory/virtualization/hyperv.t`: test cases updated with
  expected `DRIVES`, `SERIAL`, `MAC`, `IPADDRESS` and `OPERATINGSYSTEM` values
- `resources/win32/wmi/`: 9 WMI fixture files added/updated
- `Changes`: updated

## Tests

    prove -l t/tasks/inventory/virtualization/hyperv.t
    # All tests successful. Files=1, Tests=11

Perl Critic and POD syntax checks pass (`TEST_AUTHOR=1`).


Changes
```diff
@@ -7,9 +7,8 @@ Warning for developers: Don't change/remove any entry which doesn't concern your
 inventory:
 * fix #1149: Support NVMe SSD storage disks on MacOSX
 * fix #1150: Fix lxd containers inventory on linux
-* fix #1158: Fix possible use of uninitialized warning during monitor inventory
-* fix #1159: Fix possible use of uninitialized value in hash element warning during
-  windows software inventory
```
g-bougard, line 41: Please, don't erase Changes lines you don't own.

You indeed probably just need to rebase your branch, or maybe you didn't take enough care during a rebase.

lib/GLPI/Agent/Inventory.pm
```diff
@@ -100,7 +100,8 @@ my %fields = (
                             CLASS SUBCLASS NAME/ ],
     USERS            => [ qw/LOGIN DOMAIN/ ],
     VIRTUALMACHINES  => [ qw/MEMORY NAME UUID STATUS SUBSYSTEM VMTYPE VCPU
-                             MAC COMMENT OWNER SERIAL IMAGE IPADDRESS OPERATINGSYSTEM/ ],
+                             MAC COMMENT OWNER SERIAL IMAGE IPADDRESS OPERATINGSYSTEM
+                             DRIVES HOSTNAME/ ],
```
g-bougard: Here you add 2 fields which are not defined in official inventory format specs: https://github.com/glpi-project/inventory_format/blob/7fc3e1d02e959462b67350b66e6825515709ba1d/inventory.schema.json#L2008

Actually, this means any inventory providing these fields will just be refused by GLPI v10 or GLPI v11.

lib/GLPI/Agent/Task/Inventory/Virtualization/HyperV.pm
```diff
@@ -33,6 +33,25 @@ sub _getVirtualMachines {
...
+            $hostname = $obj->{DNSHostName} . '.' . $obj->{Domain};
+        } else {
+            $hostname = $obj->{DNSHostName};
+        }
+    }
+    if (!$hostname) {
+        require Sys::Hostname;
+        $hostname = Sys::Hostname::hostname();
+    }
+
```
g-bougard: Why setting host FQDN or hostname on virtualmachine ? I would better expect virtualmachine FQDN or hostname here.

The hostname of the Hyper-V running the vm should be found with the linked host.

lib/GLPI/Agent/Task/Inventory/Virtualization/HyperV.pm
```diff
@@ -71,6 +90,39 @@ sub _getVirtualMachines {
         $biosguid{$1} =~ tr/{}//d;
     }
 
+    # Index VHD sizes by file path using PowerShell Get-VHD
+    # Size is in bytes, convert to MB
+    my %vhd_size;
+    my $script = 'Get-VM | Get-VMHardDiskDrive | ForEach-Object { Get-VHD $_.Path } | Select-Object Path, Size | ForEach-Object { Write-Output ($_.Path + "|" + $_.Size) }';
+    for my $line (GLPI::Agent::Tools::Win32::runPowerShell(script => $script)) {
```
g-bougard: To be honest, using `runPowerShell()` should be used as less as possible. Actually it is only used for UWP software inventory. The problem is it can trigger secutiry tools alert.

So my question, as you also get some data via WMI, can't you access path and size via WMI ?

g-bougard (changes requested):
Hi @javierlago-ticgal 

Anyway I see few problems regarding the proposition. You can check my first comments on few code parts.

Also in your motivation, you are saying no modification is required GLPI-side, but I don't agree with that assertion. Maybe you missed them but actually the fields you're adding will make the computer inventory rejected by glpi as I said in a comment.

lib/GLPI/Agent/Inventory.pm
g-bougard: Actually, we added in the past for a customer the support of ip address and operating system. See: https://github.com/glpi-project/inventory_format/pull/146
It also involved few things to take care of:
 * handle glpi version to not send these fields if glpi server is older than the version which included the new format
 * in glpi, we also had to be sure it is nicely supported

javierlago-ticgal:
After analyzing your feedback and reviewing the GLPI source code, here is what we plan to do:

**1. Remove `HOSTNAME` entirely**
You are right, the Hyper-V host FQDN does not belong on the VM entry. We will remove it completely.

**2. Replace PowerShell with pure WMI**
We will remove the `runPowerShell`/`Get-VHD` call and obtain disk sizes directly from `MSVM_StorageAllocationSettingData` using the `VirtualQuantity` and `VirtualQuantityUnits` properties, which avoids triggering security tools.

**3. Rename `DRIVES` to `STORAGES`**
After reviewing `VirtualMachine.php` in GLPI 10.0.25 and 11.0.7, we noticed that `VMCOMPONENTS` already maps `storages` to the `Drive` class, which is the correct type for virtual hard disk files. We will rename the field accordingly in the agent and in `Inventory.pm`.

**4. Open a PR to `glpi-project/inventory_format`**
The field `storages` exists in the schema at the computer level but not inside the `virtualmachines` definition. We will open a PR to add it as an optional field — a non-breaking addition that should result in a minor version bump (1.2 → 1.3). Since GLPI already requires `^1.2`, both current releases (10.0.25 and 11.0.7) will automatically benefit from the update without any GLPI release needed.

**5. Add version gating**
We will gate the `STORAGES` field in the agent so it is only sent to GLPI servers running 10.0.25 / 11.0.7 or later, once the `inventory_format` PR is accepted.

Thanks again for guiding us through this process.

g-bougard:
Let me also speak about a possible problem. You will have to check what happens when storages are imported from the host inventory for a given vm and in the case an agent is also running on the given vm pushing computer inventory to GLPI.

The GLPI feature to create asset for vms from the host inventory can be nice but it can also introduced really bad side-effects if glpi-agent still runs in vms.

To be clear, I'm always skeptical with such features.

javierlago-ticgal:
Hi @g-bougard thank you for flagging this concern.

In our case, this change is being developed for a specific use case where the agent runs only on the Hyper-V host and there are no agents running inside the guest VMs. We are aware of the potential conflicts that could arise in mixed environments where both the host and guest have agents running, and we understand why that scenario is problematic.

Could you let us know if this limitation is a blocker for accepting the PR?

g-bougard:
If you clarify what happens in that case, this will probably not block the PR. Actually, the blocking things are the inventory_format PR has to be merged and then the code has to handle glpi server version as we does for OS and IP insertion.
The problem is more storage data has to be coherent with what a local inventory see. This is less difficult when speaking about OS or IP, but STORAGE handles more datas.
Reconciling the data remains the crux of the problem when inventory data can come from different sources.

trasher:
> [...]
> The GLPI feature to create asset for vms from the host inventory can be nice but it can also introduced really bad side-effects if glpi-agent still runs in vms.

From a GLPI point of view, I think we should just consider this option is not enabled (it's not by default, and it's not recommended). I'm not sur adding more information on VMs will make things worst.

That said, I have no idea if proposal is relevant; all that I can tell is the inventory format PR is OK for me, and can be merged if you want @g-bougard.

g-bougard:
@trasher We have to wait and see if reported data can be the same than from a local inventory.
I think I asked you to make sure virtualmachine data provided by virtualmachine host is partial, but can you confirm it is ?

trasher:
> @trasher We have to wait and see if reported data can be the same than from a local inventory. 

:+1:  I'll wait :)

> I think I asked you 

Nope.

> make sure virtualmachine data provided by virtualmachine host is partial, but can you confirm it is ?

Example in the PR is partial, yes; but that does not matter from the format pov; it only concerns GLPI.

OscarBeiro:
My apologies for the uninvited comment :)

Am I wrong or is this option in red precisely enabled to warn about this conflict?  
image

The PR is meant for the _Create computer for virtual machines_ and _create componentes_ are checked scenario. 

Currently, if you enable both and install de glpi agent in the VM, bad things happen, as the red warning states.

Let me know if I missed something.

Merci

g-bougard:
Hello @OscarBeiro 

indeed, such a warning won't prevent this can happen. And if this can happen, it will happen and indeed we still saw it happened. The discussion is more to try to limit bad side-effect as much as possible in that case.

To be honest, I would like glpi disables this feature automatically for one asset if a computer inventory is submitted, leaving a flag set somewhere (maybe on the virtualmachine list on the host asset and in a dedicated asset field) which can be unset manually if finally the agent is removed from the vm. In that case, we would even be able to remove the red warning. And I would be far less scared about such feature PR.

lib/GLPI/Agent/Task/Inventory/Virtualization/HyperV.pm
javierlago-ticgal: <p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(204, 204, 204); font-family: -apple-system, BlinkMacSystemFont, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Hi Guillaume,</p><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(204, 204, 204); font-family: -apple-system, BlinkMacSystemFont, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">After further investigation, we have confirmed through testing on a real Hyper-V host that <code style="font-family: monospace; color: rgb(215, 186, 125); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">MSVM_StorageAllocationSettingData</code> does not expose the provisioned disk size — all size-related fields (<code style="font-family: monospace; color: rgb(215, 186, 125); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">VirtualQuantity</code>, <code style="font-family: monospace; color: rgb(215, 186, 125); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">Limit</code>, <code style="font-family: monospace; color: rgb(215, 186, 125); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">AllocationUnits</code>) consistently return <code style="font-family: monospace; color: rgb(215, 186, 125); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">count</code> regardless of the actual disk size. <code style="font-family: monospace; color: rgb(215, 186, 125); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">MSVM_VirtualHardDiskSettingData</code> and <code style="font-family: monospace; color: rgb(215, 186, 125); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">MSVM_VirtualHardDiskState</code> return no data at all.</p><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(204, 204, 204); font-family: -apple-system, BlinkMacSystemFont, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">The only way to retrieve the provisioned size is through <code style="font-family: monospace; color: rgb(215, 186, 125); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">Get-VHD</code>, which is a read-only cmdlet from the official Hyper-V management module. To address your security concern, we have reduced the PowerShell usage to the absolute minimum — a single call only to retrieve the disk size, while all other data (disk paths, VM identity) still comes from WMI:</p><div class="codeBlockWrapper_-a7MRw" style="position: relative; margin: 8px 0px; color: rgb(204, 204, 204); font-family: -apple-system, BlinkMacSystemFont, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><button class="copyButton_CEmTFw copyButton_-a7MRw" title="Copy code" aria-label="Copy code to clipboard" style="color: rgb(204, 204, 204); font-family: -apple-system, BlinkMacSystemFont, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; background: none 0% 0% / auto repeat scroll padding-box border-box rgb(40, 44, 52); border-color: rgb(58, 63, 75); border-style: solid; border-width: 1px; border-image: none 100% / 1 / 0 stretch; cursor: pointer; opacity: 0; display: flex; border-radius: 4px; justify-content: center; align-items: center; padding: 4px; transition: opacity 0.15s, background 0.15s; position: absolute; top: 4px; right: 4px;"><svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20" fill="currentColor" aria-hidden="true" data-slot="icon" class="copyIcon_CEmTFw"><path fill-rule="evenodd" d="M15.988 3.012A2.25 2.25 0 0 1 18 5.25v6.5A2.25 2.25 0 0 1 15.75 14H13.5v-3.379a3 3 0 0 0-.879-2.121l-3.12-3.121a3 3 0 0 0-1.402-.791 2.252 2.252 0 0 1 1.913-1.576A2.25 2.25 0 0 1 12.25 1h1.5a2.25 2.25 0 0 1 2.238 2.012ZM11.5 3.25a.75.75 0 0 1 .75-.75h1.5a.75.75 0 0 1 .75.75v.25h-3v-.25Z" clip-rule="evenodd"></path><path d="M3.5 6A1.5 1.5 0 0 0 2 7.5v9A1.5 1.5 0 0 0 3.5 18h7a1.5 1.5 0 0 0 1.5-1.5v-5.879a1.5 1.5 0 0 0-.44-1.06L8.44 6.439A1.5 1.5 0 0 0 7.378 6H3.5Z"></path></svg></button><pre style="overflow-x: auto; white-space: pre; box-sizing: border-box; border-radius: 4px; max-width: 100%; margin: 0px; padding: 8px;"><code class="language-perl" style="font-family: monospace; color: rgb(215, 186, 125); background-color: rgba(255, 255, 255, 0.1); padding: 0px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">my $ps_script = 'Get-VHD -Path "' . $path . '" | Select-Object -ExpandProperty Size';
my $size_bytes = runPowerShell(script =&gt; $ps_script);
my $size_mb = $size_bytes ? int($size_bytes / 1024 / 1024) : 0;
</code></pre></div><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(204, 204, 204); font-family: -apple-system, BlinkMacSystemFont, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">We tested this on a real Hyper-V host with dynamic disks of different sizes, including a disk on a network share (UNC path), and the results were correct in all cases:</p>
Disk | Provisioned size
-- | --
C:\HyperV\vm1.vhdx | ~9.4 GB
C:\HyperV\vm2.vhdx | ~4.5 GB
\\server\share\disk.vhdx | 2 GB

<p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(204, 204, 204); font-family: -apple-system, BlinkMacSystemFont, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">WMI returns <code style="font-family: monospace; color: rgb(215, 186, 125); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">FileSize</code> (physical size on disk) for dynamic disks, which is always much smaller than the provisioned size and would produce incorrect inventory data. For this reason, the use of <code style="font-family: monospace; color: rgb(215, 186, 125); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">Get-VHD</code> is necessary to ensure the inventory accurately reflects the configured size of each disk.</p><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(204, 204, 204); font-family: -apple-system, BlinkMacSystemFont, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Thank you</p>

g-bougard:
Hi @javierlago-ticgal 

in your update, you should also probably try to add ip and os as the format supports them [since GLPI 10.0.17 in ESX inventory](https://github.com/glpi-project/glpi-agent/blob/e9c47c5442be4954ebe53859177306487c11b4bf/lib/GLPI/Agent/SOAP/VMware/Host.pm#L430). This is definitively not a requirement to accept your PR, just a suggestion to enhance Hyper-V VM inventory. If this is too difficult, just ignore my suggestion.

javierlago-ticgal:
Hi @g-bougard ,

Thank you for the suggestion. We have investigated the feasibility and both fields appear to be accessible via WMI when Hyper-V Integration Services are installed on the guest VMs — MSVM_SummaryInformation for IP and MSVM_KvpExchangeDataItem for OS, following the same pattern as the ESX implementation.

However, our current test environment does not have fully configured VMs with Integration Services running, so we cannot confirm this reliably yet. We are setting up a proper environment to validate it. If the results are positive we will include these fields in this PR, otherwise we will open a follow-up PR.

Thank you

javierlago-ticgal:
Hi @g-bougard 

Here is a summary of the current state of this update:

**DRIVES instead of STORAGES**

After reviewing `VirtualMachine.php` more carefully, we found that GLPI maps `storages → Drive` (Components tab, physical hardware) and `drives → Volume` (Volumes tab). Since VHD/VHDX files are logical volumes, not physical drives, `drives` is the semantically correct field. We have updated the agent, `Inventory.pm`, the test fixtures, and the `inventory_format` PR accordingly.

**IPADDRESS and OPERATINGSYSTEM via KVP**

Following your suggestion, we implemented both fields using `Msvm_KvpExchangeComponent`, which exposes guest data provided by Hyper-V Integration Services without requiring PowerShell. The agent reads `NetworkAddressIPv4`, `OSName` and `OSVersion` from the KVP exchange items and maps them to `IPADDRESS` and `OPERATINGSYSTEM`. Both fields are only populated for running VMs with Integration Services installed, which is the expected behavior.

**Version gating**

All extended fields (`DRIVES`, `IPADDRESS`, `OPERATINGSYSTEM`) are gated behind `supportsGlpiVersion('10.0.25')` and are only sent when the server reports a compatible version.

**Test results on real GLPI 10.0.25**

We validated against a real GLPI 10.0.25 instance with three Hyper-V VMs:
- Virtual disks appear correctly in the **Volumes** tab of each VM asset
- OS (`Ubuntu 6.8.0`) is visible on the running VM
- Disk sizes are accurate (local VHDs and UNC network paths)


We will wait for the `inventory_format` PR to be merged before this one is considered ready.

Thank you

lib/GLPI/Agent/Task/Inventory/Virtualization/HyperV.pm
```diff
@@ -19,20 +19,32 @@ sub doInventory {
     my (%params) = @_;
 
     my $inventory = $params{inventory};
+    my $logger    = $params{logger};
 
-    foreach my $machine (_getVirtualMachines()) {
+    foreach my $machine (_getVirtualMachines($inventory, $logger)) {
```
g-bougard: Here, we have a convention where you should better pass %params and reuse `my (%params) = @_;`
```suggestion
    foreach my $machine (_getVirtualMachines(%params)) {
```

lib/GLPI/Agent/Task/Inventory/Virtualization/HyperV.pm
```diff
@@ -19,20 +19,32 @@ sub doInventory {
...
         $inventory->addEntry(
             section => 'VIRTUALMACHINES', entry => $machine
         );
     }
 }
 
 sub _getVirtualMachines {
+    my ($inventory, $logger) = @_;
```
g-bougard: To follow my previous comment:
```suggestion
    my (%params) = @_;

    my $logger = $params{logger};
```

lib/GLPI/Agent/Task/Inventory/Virtualization/HyperV.pm
```diff
@@ -19,20 +19,32 @@ sub doInventory {
...
 
     GLPI::Agent::Tools::Win32->require();
 
     my @machines;
 
+    # Determine once whether GLPI supports extended VM fields
+    my $extended = $inventory && $inventory->supportsGlpiVersion('10.0.25');
```
g-bougard: The `$extended` value is not correct:
1. Operating system and ip address fields are supported since v10.0.25
2. drive is still not supported, it will be from the GLPI 11 or 12 version in which your inventory_format will be included

So you'll have to handle $extended value in this way:
```suggestion
    my $inventory = $params{inventory};
    my $extended = 0;
    if ($inventory) {
        $extended = $inventory->supportsGlpiVersion('12') ? 2 :
            $inventory->supportsGlpiVersion('10.0.25') ? 1 : 0;
    }
```
Then you'll have to test `$extended > 1` at the time you want to add drives inventory.

For the `2` value, you'll have to fix it after inventory_format PR will be merged and when you know which glpi version will integrate it.

lib/GLPI/Agent/Task/Inventory/Virtualization/HyperV.pm
```diff
@@ -19,20 +19,32 @@ sub doInventory {
...
+    if ($extended) {
+        $logger->debug("Hyper-V: GLPI supports extended VM fields (DRIVES, IPADDRESS, OPERATINGSYSTEM)")
+            if $logger;
+    } else {
+        $logger->debug("Hyper-V: GLPI version does not support extended VM fields (requires 10.0.25+), collecting basic inventory only")
+            if $logger;
+    }
```
g-bougard: You'll have to handle `$extended` value.

lib/GLPI/Agent/Task/Inventory/Virtualization/HyperV.pm
```diff
@@ -71,6 +83,88 @@ sub _getVirtualMachines {
...
+            my $id = $object->{InstanceID} // '';
+            next unless $id =~ /^Microsoft:([^\\]+)/;
+            my $vm_guid = $1;
+            my $path = ref($object->{HostResource}) eq 'ARRAY'
+                         ? $object->{HostResource}[0]
+                         : $object->{HostResource};
+
+            # Skip ISO images — Get-VHD does not support them
+            if ($path =~ /\.iso$/i) {
+                $logger->debug("Hyper-V: skipping ISO image '$path'")
```
g-bougard: Use better debug2 level here.
```suggestion
                $logger->debug2("Hyper-V: skipping ISO image '$path'")
```

lib/GLPI/Agent/Task/Inventory/Virtualization/HyperV.pm
```diff
@@ -71,6 +83,88 @@ sub _getVirtualMachines {
...
+                    if $logger;
+                next;
+            }
+
+            my ($size_bytes) = GLPI::Agent::Tools::Win32::runPowerShell(
+                script => 'Get-VHD -Path "' . $path . '" | Select-Object -ExpandProperty Size'
+            );
```
g-bougard, line 176: Is there no way to use WMI request here in place of `runPowerShell()` api ?

lib/GLPI/Agent/Task/Inventory/Virtualization/HyperV.pm
```diff
@@ -71,6 +83,88 @@ sub _getVirtualMachines {
         $biosguid{$1} =~ tr/{}//d;
     }
 
+    my %drives;
+    my %kvp;
+    if ($extended) {
```
g-bougard: To follow `$extended` value:
```suggestion
    if ($extended > 1) {
```

lib/GLPI/Agent/Task/Inventory/Virtualization/HyperV.pm
```diff
@@ -71,6 +83,88 @@ sub _getVirtualMachines {
...
+                $logger->debug2("Hyper-V: VHD '$path' size = $size_bytes bytes")
+                    if $logger;
+            }
+            my $size_mb = $size_bytes ? int($size_bytes / 1024 / 1024) : 0;
+            push @{$drives{$vm_guid}}, {
+                VOLUMN => $path,
+                TOTAL  => $size_mb,
+            };
+        }
+
```
g-bougard: To handle `$extended` value:
```suggestion
        }
    }

    if ($extended) {
```

lib/GLPI/Agent/Task/Inventory/Virtualization/HyperV.pm
```diff
@@ -103,10 +201,37 @@ sub _getVirtualMachines {
             VCPU      => $vcpu{$object->{Name}},
         };
 
-        push @machines, $machine;
+        if ($extended) {
```
g-bougard, line 308: Here you need to handle `$extended` value.

Changes
```diff
@@ -10,6 +10,8 @@ inventory:
 * fix #1158: Fix possible use of uninitialized warning during monitor inventory
 * fix #1159: Fix possible use of uninitialized value in hash element warning during
   windows software inventory
+* Add STORAGES support for Hyper-V VMs: collect virtual hard disk paths and
```
g-bougard: Update the comment as the PR inventory DRIVES in place of STORAGES

lib/GLPI/Agent/Inventory.pm
```diff
@@ -402,8 +409,8 @@ sub addEntry {
             delete $entry->{$field};
             next;
         }
-        # sanitize value
-        my $value = getSanitizedString($entry->{$field});
+        # sanitize value (skip refs like arrayrefs/hashrefs)
+        my $value = ref($entry->{$field}) ? $entry->{$field} : getSanitizedString($entry->{$field});
```
g-bougard: Here, we should do better. But I think I should handle this part as we are touching glpi-agent core.

I'll prepare a dedicated PR for that and I'll ask you to rebase your PR on it.

g-bougard (changes requested):
Hi @javierlago-ticgal 

thank you for the updates.

I think there's some changes required. Can you check my comments ?

I'll prepare a PR for the glpi-agent core changes as we need to do better and I'll ask you to rebase your PR on that PR.

javierlago-ticgal:
Hi @g-bougard,

Regarding the `DRIVES` field gating at GLPI >= 12: we understand the
requirement to wait for the `inventory_format` PR to be merged first.

We have already opened [glpi-project/inventory_format#170](https://github.com/glpi-project/inventory_format/pull/170) to add `drives`
as an optional field in the `virtualmachines` schema. 

Our concern is that our clients are currently running GLPI 10.0.25 (the
latest stable 10.x release). If `drives` support lands only in GLPI 11
or 12, they would not be able to use this feature until they upgrade.

Could you clarify whether, once glpi-project/inventory_format#170 is
merged, a backport to a GLPI 10.x patch release would be feasible? Or
should we expect this to be available only from GLPI 11/12 onwards?

This will help us set the correct version gate — either keeping
`$extended >= 2` or lowering it to `$extended >= 1` (10.0.25) once
the schema is official.

In the meantime, we plan to maintain a separate branch for our client
with `DRIVES` enabled at the 10.0.25 level. Once our client migrates
to the version that officially includes this support, they will switch
back to the official agent release.

Thank you

lib/GLPI/Agent/Task/Inventory/Virtualization/HyperV.pm
javierlago-ticgal: Hi @g-bougard,

We investigated all available WMI classes that could provide the
provisioned disk size without using PowerShell:

- **`MSVM_StorageAllocationSettingData`**: provides `VirtualQuantity`
  and `VirtualQuantityUnits`, but these fields return the number of
  storage blocks, not the actual provisioned size in bytes. In our
  tests on a real Hyper-V host the values were inconsistent and did
  not match the real disk size.

- **`MSVM_VirtualHardDiskSettingData`**: returns `MaxInternalSize`
  (the provisioned size), but only when queried from inside the guest
  VM — not accessible from the host.

- **`MSVM_VirtualHardDiskState`**: returns `FileSize` (physical size
  on disk), which for dynamic disks is always much smaller than the
  provisioned size and would produce incorrect inventory data.

`Get-VHD` is a read-only cmdlet from the official Hyper-V PowerShell
management module and is the only reliable way to retrieve the
provisioned size from the host. We have reduced its usage to the
absolute minimum — a single call per disk only to retrieve the size,
while all other data (disk paths, VM identity) still comes from WMI.

Thank you

g-bougard:
Hi @javierlago-ticgal 

I still can't tell you in which GLPI versions the updated format could be integrated. We will see that point later. If it is integrated in future 10 and 11 versions, you will have to handle the cases v11.0.0 to at least v11.0.7 won't support the format.

Anyway, you can force glpi-agent to support features from future glpi versions using the [glpi-version option](https://glpi-agent.readthedocs.io/en/latest/configuration.html#glpi-version). Of course, the features must really supported server-side with a dedicated patch and the updated format should also have been pushed to glpi server or the inventory may be refused or even worse involves errors.

So if you maintain a dedicated branch for your customer, you'll only have to configure concerned agents with expected glpi-version value. And it will only concern servers managing Hyper-V virtualmachines. If finally, the format is only merged in glpi v12, you'll have to set : `glpi-version = 12`

Also you said in your previous comment:
> We will wait for the inventory_format PR to be merged before this one is considered ready.

Just to clear, inventory_format PR will be merged only if this one is ready and accepted before.

javierlago-ticgal:
Hi @g-bougard,

We have pushed the latest changes according to your suggestions:

- `_getVirtualMachines()` now uses the `%params` convention
- `$extended` split into three levels (0, 1, 2) as suggested
- `DRIVES` gated behind `$extended > 1` (GLPI >= 12 placeholder)
- `IPADDRESS` and `OPERATINGSYSTEM` remain at `$extended >= 1` (GLPI >= 10.0.25)
- ISO skip log changed to `debug2` level
- `Changes` entry updated: STORAGES -> DRIVES

Regarding our client running GLPI 10.0.25: after reviewing the
documentation we found the `glpi-version` option (available since
GLPI Agent v1.12). We will instruct our technicians to set the
following in the agent configuration on Hyper-V hosts:

```ini
glpi-version = 12
```

This avoids the need to maintain a separate branch. Once the
`inventory_format` PR is merged and you confirm the actual GLPI
version that will include it, we will update the placeholder value
accordingly.

Thank you for your guidance throughout this process.

g-bougard:
Can you still rebase your PR against #1168 ?

You could then include this change in Inventory.pm so DRIVES entries will be checked too:
```patch
diff --git a/lib/GLPI/Agent/Inventory.pm b/lib/GLPI/Agent/Inventory.pm
index 4597e11de..c70863c76 100644
--- a/lib/GLPI/Agent/Inventory.pm
+++ b/lib/GLPI/Agent/Inventory.pm
@@ -120,6 +120,10 @@ my %checks = (
         }
     },
     VIRTUALMACHINES => {
+        DRIVES => {
+            not_before  => glpiVersion('12'),
+            as_ref      => 'DRIVES',
+        },
         IPADDRESS => {
             not_before  => glpiVersion('10.0.17'),
         },
```

javierlago-ticgal:
Hi @g-bougard,

We have rebased our PR on top of `feature/extended-inventory-check` and
applied the suggested change to `Inventory.pm`:

- Added `DRIVES` entry in `%checks` for `VIRTUALMACHINES` with
  `not_before => glpiVersion('12')` and `as_ref => 'DRIVES'`
- Removed our previous manual `ref()` check in `addEntry` as it is now
  properly handled by your `as_ref` mechanism

All tests pass locally.

Thank you for your guidance and patience throughout this review process.
We really appreciate the time you have taken to help us align with the
project conventions and ensure the implementation is done correctly.

We will wait for `feature/extended-inventory-check` to be merged before
requesting a final review.

javierlago-ticgal:
Hi @g-bougard,

Following your suggestion to use `glpi-version = 12` for our customer's environment, we ran QA tests on two real Hyper-V hosts. The results revealed that `glpi-version` does **not** override the server-reported version when the agent connects to a native GLPI server.

**Root cause (found in `Target/Server.pm`)**

`getTaskVersion()` returns the server-reported version whenever `_server_task_support` is populated by the contact response:

    sub getTaskVersion {
        my ($self, $task) = @_;
        $task = lc($task);

        return $self->{_glpi} unless $task && $self->{_server_task_support} && $self->{_server_task_support}->{$task};
        return $self->{_server_task_support}->{$task}->{version} // $self->{_glpi} // '';
    }

When the agent contacts a GLPI 10.0.25 server, the server responds with `"version":"10.0.25"` in the contact JSON, `_server_task_support->{inventory}` gets populated, and from that point `_glpi` (the `glpi-version` config value) is completely ignored. Our QA logs confirmed this: even with `glpi-version = 12` set in `agent.cfg`, `$extended` remained at `1` and DRIVES were never included.

**What this means for our customer**

Since `glpi-version` does not act as an override, our customer cannot activate DRIVES with configuration alone. We have created a dedicated branch where we set both gates to `glpiVersion('10.0.25')` (matching their patched server), and they will switch back to the official agent once the feature lands in a supported GLPI release.

**Upstream PR — no changes**

We are keeping the upstream PR exactly as agreed: `$extended > 1` for DRIVES in `HyperV.pm` and `not_before => glpiVersion('12')` in `Inventory.pm`. We will update the placeholder version once you confirm which GLPI release will include the `inventory_format` changes.

**Separate issue?**

Would it make sense to open a separate issue/PR to fix `getTaskVersion()` so that an explicitly set `glpi-version` takes priority over the server-reported version? That would make the option behave as documented and as you described it.

Thank you.

javierlago-ticgal:
Hi @g-bougard,

We pushed two commits to `develop` after further QA validation on a real Hyper-V host with DRIVES enabled (GLPI 10.0.25):

**1. KVP `NetworkAddressIPv4` deprecation fallback**

On hosts with newer Hyper-V Integration Services installed in some guest VMs, `Msvm_KvpExchangeComponent`'s `NetworkAddressIPv4` key returns a deprecation notice string instead of an actual IP address. Storing that text as `IPADDRESS` caused GLPI to reject the inventory.

We now validate the value with an IP regex before storing it. If it doesn't look like an IP, we flag the VM and fall back to `Msvm_GuestNetworkAdapterConfiguration` (available since Windows Server 2012 R2) to retrieve the IP instead. The fallback only runs when at least one VM needs it, so hosts without the deprecation are unaffected.

**2. `_checkEntry()` returning `false` instead of the validated entry**

Separately, we hit a new **500 Internal Server Error** on submission:

    property_exists(): Argument #1 ($object_or_class) must be of type object|string, bool given

The generated JSON showed every `DRIVES` entry as `false` instead of `{"volumn": "...", "total": ...}`:

    "drives": [
       false
    ],

Root cause: `_checkEntry()` in `Inventory.pm` (added in b493066) has no explicit `return $entry;`. Since the sub's last statement is a bare `foreach` loop, it falls off the end and returns Perl's internal empty/false scalar instead of the validated hashref. Back in `addEntry()`:

    my $val = $self->_checkEntry($fieldentry, $ref);
    push @{$value}, $val if defined($val);

pushes that scalar instead of the hashref, since it's still `defined`. We added the missing `return $entry;`. Verified locally that DRIVES now serializes correctly, and `prove -l t/tasks/inventory/virtualization/hyperv.t` passes (7/7).

Both fixes are independent of the `glpi-version` issue we reported previously — that one remains open and we haven't touched the `$extended > 1` / `glpiVersion('12')` gate in this PR.

Thank you.

.claude/guia-pull-request.md
g-bougard, line 1: Please remove this file as you should added it by mistake.

lib/GLPI/Agent/Task/Inventory/Virtualization/HyperV.pm
```diff
@@ -71,6 +94,122 @@ sub _getVirtualMachines {
...
+    }
+    if ($extended) {
+        foreach my $object (GLPI::Agent::Tools::Win32::getWMIObjects(
+            moniker    => 'winmgmts://./root/virtualization/v2',
+            altmoniker => 'winmgmts://./root/virtualization',
+            class      => 'Msvm_KvpExchangeComponent',
+            properties => [ qw/SystemName GuestIntrinsicExchangeItems/ ]
+        )) {
+            my $vm_guid = $object->{SystemName} // next;
+            my $items   = $object->{GuestIntrinsicExchangeItems} // next;
```
g-bougard: This code is not correct.

If `$object->{SystemName}` can be zero and the test is only to check values are defined, use this code:
```suggestion
            my $vm_guid = $object->{SystemName}
                or next;
            my $items   = $object->{GuestIntrinsicExchangeItems}
                or next;
```
Here `or` is the operator to use as it's the one with the least priority.

lib/GLPI/Agent/Task/Inventory/Virtualization/HyperV.pm
```diff
@@ -71,6 +94,122 @@ sub _getVirtualMachines {
...
+            $items = [$items] unless ref($items) eq 'ARRAY';
+            foreach my $xml (@$items) {
+                $xml =~ s/&quot;/"/g;
+                my ($name) = $xml =~ m{<PROPERTY NAME="Name"[^>]*><VALUE>([^<]*)</VALUE>};
+                my ($data) = $xml =~ m{<PROPERTY NAME="Data"[^>]*><VALUE>([^<]*)</VALUE>};
+                next unless defined $name && defined $data && length($data);
```
g-bougard: Here you can use our `empty()` api which still checks if a value is defined and has a length.
```suggestion
                next unless empty($name) && empty($data);
```

lib/GLPI/Agent/Task/Inventory/Virtualization/HyperV.pm
```diff
@@ -71,6 +94,122 @@ sub _getVirtualMachines {
...
+        if ($need_fallback) {
+            $logger->debug("Hyper-V: KVP IP keys deprecated, falling back to Msvm_GuestNetworkAdapterConfiguration")
+                if $logger;
+            foreach my $object (GLPI::Agent::Tools::Win32::getWMIObjects(
+                moniker    => 'winmgmts://./root/virtualization/v2',
+                class      => 'Msvm_GuestNetworkAdapterConfiguration',
+                properties => [ qw/InstanceID IPAddresses/ ]
+            )) {
+                my $instance_id = $object->{InstanceID} // next;
+                my $ips         = $object->{IPAddresses} // next;
```
g-bougard: As before, there's a wrong syntax here, better use:
```suggestion
                my $instance_id = $object->{InstanceID}
                    or next;
                my $ips         = $object->{IPAddresses}
                    or next;
```

lib/GLPI/Agent/Task/Inventory/Virtualization/HyperV.pm
```diff
@@ -103,10 +247,39 @@ sub _getVirtualMachines {
...
+            $machine->{DRIVES} = $drives{$object->{Name}} // [];
+        }
+        if ($extended) {
+            my $vm_kvp = $kvp{$object->{Name}};
+            if ($vm_kvp) {
+                $machine->{IPADDRESS} = $vm_kvp->{IPADDRESS}
+                    if defined $vm_kvp->{IPADDRESS};
+                if ($vm_kvp->{OSName} || $vm_kvp->{OSVersion}) {
+                    my $full_name = join(' ',
+                        grep { defined $_ && length $_ }
```
g-bougard: Use our `empty()` api here:
```suggestion
                        grep { !empty($_) }
```

g-bougard (changes requested):
Hi @javierlago-ticgal 

I see few things to update. Can you check my comments ?

Also, it seems there are merge conflicts. Can you rebase your PR to latest develop branch update ?

javierlago-ticgal:
Hi @g-bougard, thanks for the review! All changes have been applied and the branch has been rebased on latest develop.

**Changes applied:**
- `// next` replaced with `or next` for `SystemName`, `GuestIntrinsicExchangeItems`, `InstanceID` and `IPAddresses` guards
- `grep { defined $_ && length $_ }` replaced with `grep { !empty($_) }`
- Removed the accidentally committed `.claude/guia-pull-request.md` file

**Regarding the `empty()` suggestion for the `next unless` guard:**

I noticed the suggested change had the logic inverted. `empty()` returns true when the value is undefined or has no length, so:

```perl
# suggested — skips only when BOTH are empty (wrong)
next unless empty($name) && empty($data);

# applied — skips when either is empty (correct)
next if empty($name) || empty($data);

File: lib/GLPI/Agent/Task/Inventory/Virtualization/HyperV.pm [hunk 20]
```diff
@@ -71,6 +94,126 @@ sub _getVirtualMachines {
...
+            # Skip file types unsupported by Get-VHD (ISOs, virtual floppy disks)
+            if ($path =~ /\.(?:iso|vfd)$/i) {
+                $logger->debug2("Hyper-V: skipping unsupported file type '$path'")
+                    if $logger;
+                next;
+            }
+
+            my ($size_bytes) = GLPI::Agent::Tools::Win32::runPowerShell(
+                script => 'Get-VHD -Path "' . $path . '" | Select-Object -ExpandProperty Size'
+            );
```
g-bougard, line 176: I noticed a little problem here. In runPowerShell() api, we use a filename template dedicated to the only place this api was used before. Can you add an update on this api to support providing a template ? Like this:
```perl
    my $fh = File::Temp->new(
        TEMPLATE    => $params{TEMPLATE} // 'get-appxpackage-XXXXXX',
        SUFFIX      => '.ps1'
    );
```
And update here to set the TEMPLATE like:
```suggestion
            my ($size_bytes) = GLPI::Agent::Tools::Win32::runPowerShell(
                script      => 'Get-VHD -Path "' . $path . '" | Select-Object -ExpandProperty Size',
                TEMPLATE    => 'get-vhd-size-XXXXXX',
            );
```

lib/GLPI/Agent/Task/Inventory/Virtualization/HyperV.pm
```diff
@@ -103,10 +251,39 @@ sub _getVirtualMachines {
             VCPU      => $vcpu{$object->{Name}},
         };
 
-        push @machines, $machine;
+        if ($extended > 1) {
+            $machine->{DRIVES} = $drives{$object->{Name}} // [];
+        }
```
g-bougard: I noticed you had tests where DRIVES was an empty list. You should just not set it if list is empty:
```suggestion
        if ($extended > 1 && $drives{$object->{Name}} && @{$drives{$object->{Name}}}) {
            $machine->{DRIVES} = $drives{$object->{Name}};
        }
```
You'll have also to fix some tests.

Changes
g-bougard: Again you made the same mistake during the rebase.

Please keep your attention on this problem.

g-bougard: Also, move your entry into the "inventory" section of v1.19 release.

g-bougard (changes requested):
Hi @javierlago-ticgal 

thank you for the updates. And yes, I was wrong with the test using `empty()` api, good shot.

I see few other more things to change. Can you take care on it ?

g-bougard:
Also, I just merged #1168 into develop branch with your own update to return checked value.

Can you rebase again your branch against develop ?

Thank you

lib/GLPI/Agent/Task/Inventory/Virtualization/HyperV.pm
```diff
@@ -71,6 +94,127 @@ sub _getVirtualMachines {
...
+            if ($path =~ /\.(?:iso|vfd)$/i) {
+                $logger->debug2("Hyper-V: skipping unsupported file type '$path'")
+                    if $logger;
+                next;
+            }
+
+            my ($size_bytes) = GLPI::Agent::Tools::Win32::runPowerShell(
+                script   => 'Get-VHD -Path "' . $path . '" | Select-Object -ExpandProperty Size',
+                TEMPLATE => 'get-vhd-size-XXXXXX',
+            );
```
g-bougard, line 176: 2 other questions about this code:
1. Can you add `logger   => $logger` as last parameter so the finally run command is logged ?
2. I've a security concern also with `$path` variable which is a windows path inserted. Is it possible to craft it so it can contain a double-quote to corrupt the generated powershell script or is it possible to insert any other escaping character ? The question is essentially to know if a non authorized user can force glpi-agent to run unexpected code. If this is possible, you have to handle all possible cases.

g-bougard (changes requested):
Hi @javierlago-ticgal 

thank you for the last updates.

I've again some concern with this runPowerShell() api usage. Can you check my comment below ?

lib/GLPI/Agent/Task/Inventory/Virtualization/HyperV.pm
javierlago-ticgal: Hi @g-bougard,

Thanks for the review feedback — addressed both points:

**1. `logger => $logger`** — added as the last parameter to `runPowerShell` so the executed command appears in debug output.

**2. Path injection concern** — analyzed the attack surface:

- `$path` comes from `MSVM_StorageAllocationSettingData.HostResource` (WMI, Hyper-V admin only), so an unprivileged user cannot inject data there.
- However, `"` double-quotes are illegal in NTFS paths, so that specific vector does not exist in practice.
- **Real issue found**: `$` *is* valid in Windows paths — UNC admin shares like `\\server\C$\VMs\vm.vhdx` are common in Hyper-V environments. PowerShell expands `$` in double-quoted strings, which would silently corrupt the generated script.

**Fix applied**: switched to single-quoted PowerShell strings (no variable expansion, no escape sequences) and added `s/'/''/g` to escape any literal single-quotes in the path per PowerShell convention. Updated the test mock and added a `C$`-path fixture case to cover this explicitly.

---

Separately, while validating this PR on our test environment replicating the customer's setup, we identified an additional need: their GLPI rules match VMs using **UUID + serial number** to determine whether a VM already exists in the inventory. Without the serial number (and MAC as a secondary identifier), the rule can fail to match and generate **duplicate VM entries** in GLPI.

We have implemented two additional fields in `HyperV.pm` on our side:

- **`SERIAL`** — collected from `MSVM_VirtualSystemSettingData.BIOSSerialNumber` (extends the existing query we already do for `BIOSGUID`).
- **`MAC`** — first MAC address from `Msvm_SyntheticEthernetPortSettingData`, with a fallback to `Msvm_EmulatedEthernetPortSettingData` for Generation 1 legacy VMs. Formatted as `XX:XX:XX:XX:XX:XX`.

Neither field requires a GLPI version gate — both `SERIAL` and `MAC` are basic `VIRTUALMACHINES` fields with no `not_before` constraint in `Inventory.pm`. Validated in our test environment: serial numbers are correctly retrieved and MACs are present on all VMs that have a configured network adapter.

**Would you prefer we include these changes in this PR** (since it already covers extended Hyper-V VM inventory beyond just DRIVES — STATUS, IPADDRESS, OPERATINGSYSTEM), **or would you rather we open a follow-up PR** to keep this one focused?

Happy to go either way, just wanted your input before adding more changes mid-review.

Thanks!

g-bougard: You can add your new code here but you should probably update the PR title and don't forget to update the Changes file to include this feature.

lib/GLPI/Agent/Task/Inventory/Virtualization/HyperV.pm
```diff
@@ -82,16 +82,59 @@ sub _getVirtualMachines {
...
+        )) {
+            my $id = $object->{InstanceID}
+                or next;
+            next unless $id =~ /^Microsoft:([^\\]+)/;
+            my $vm_guid = $1;
+            next if defined $mac{$vm_guid};
+            my $addr = $object->{Address}
+                or next;
+            $addr =~ s/(..)(?=.)/$1:/g;
+            $mac{$vm_guid} = uc($addr);
```
g-bougard: Here if I understand well, you want to format a mac address.

You should better use `alt2canonical()` api from our GLPI::Agent::Tools::Network module. You'll have to add a `use` entry for it.

Also why forcing it to be in upper case ? MAC address standard is more lower case imho. If their is a good reason to force it in upper case, please add a comment.

lib/GLPI/Agent/Task/Inventory/Virtualization/HyperV.pm
```diff
@@ -82,16 +82,59 @@ sub _getVirtualMachines {
...
+    )) {
+        my $id = $object->{InstanceID}
+            or next;
+        next unless $id =~ /^Microsoft:([^\\]+)/;
+        my $vm_guid = $1;
+        next if defined $mac{$vm_guid};
+        my $addr = $object->{Address}
+            or next;
+        $addr =~ s/(..)(?=.)/$1:/g;
+        $mac{$vm_guid} = uc($addr);
```
g-bougard: See my comment below for the fallback case.

g-bougard (changes requested):
Hi,

thank you for the addition.

Here I only have a concern with the mac address formatting. Check my comment below. But I'm okay with that addition.

g-bougard:
I forgot also to say you need to rebase your PR as there's a conflict on the Changes file after an entry was added yesterday.

g-bougard:
Something seems have broken these 2 tests:
* t/tasks/esx.t
* t/tasks/inventory/virtualization/virtualbox.t

g-bougard:
It seems you removed MAC in 4c1c7a4, here:
```diff
diff --git a/lib/GLPI/Agent/Inventory.pm b/lib/GLPI/Agent/Inventory.pm
index 73ed7c2bb..6a988614b 100644
--- a/lib/GLPI/Agent/Inventory.pm
+++ b/lib/GLPI/Agent/Inventory.pm
@@ -100,8 +100,8 @@ my %fields = (
                             CLASS SUBCLASS NAME/ ],
     USERS            => [ qw/LOGIN DOMAIN/ ],
     VIRTUALMACHINES  => [ qw/MEMORY NAME UUID STATUS SUBSYSTEM VMTYPE VCPU
-                             MAC COMMENT OWNER SERIAL IMAGE IPADDRESS OPERATINGSYSTEM
-                             DRIVES/ ],
+                             COMMENT OWNER SERIAL IMAGE IPADDRESS OPERATINGSYSTEM
+                             DRIVES NETWORKS/ ],
     VOLUME_GROUPS    => [ qw/VG_NAME PV_COUNT LV_COUNT ATTR SIZE FREE VG_UUID
                              VG_EXTENT_SIZE/ ],
     VERSIONPROVIDER  => [ qw/NAME VERSION COMMENTS PERL_EXE PERL_VERSION PERL_ARGS

```

Is there a reason or is this a mistake ?

javierlago-ticgal:
Hi @g-bougard ,

You are right, it was a mistake. When we added the NETWORKS field we accidentally
removed MAC, which is still used by the ESX and VirtualBox (and Virtuozzo) modules.
This caused t/tasks/esx.t and t/tasks/inventory/virtualization/virtuozzo.t to crash
mid-run with "Bad plan" errors — addEntry was failing on the unrecognised MAC field
before all planned tests could complete.

We have restored MAC so both fields now coexist:

    VIRTUALMACHINES => [ qw/MEMORY NAME UUID STATUS SUBSYSTEM VMTYPE VCPU
                             MAC COMMENT OWNER SERIAL IMAGE IPADDRESS OPERATINGSYSTEM
                             DRIVES NETWORKS/ ],

All three tests (esx.t, virtualbox.t, hyperv.t) pass now. Sorry for the oversight.

g-bougard (approved):

g-bougard:
Hi @javierlago-ticgal 

okay, so I merged your PR.

I still need a confirmation GLPI 12 will support it. So comment this PR when you have your GLPI code merged for GLPI 12. If I don't have this confirmation, we may have to remove the code before GLPI 12 is released.

PR #1170, opened 2026-06-02
fix: update macOS scripts to use modern launchctl API (Ventura and later)

CristianPizaTR:
## Problem

On macOS 13 Ventura, 14 Sonoma, 15 Sequoia and 26 Tahoe, the `.pkg` installer fails to start the GLPI Agent service after installation. The `postinstall` script calls `launchctl load`, which was deprecated in macOS 10.10 and now returns **`Load failed: 5: Input/output error`** on modern macOS versions.

The same issue affects the `preinstall` script (upgrade scenario) and `uninstaller.sh`, both of which call `launchctl unload`.

Additionally, the LaunchDaemon plist generated by `postinstall` is missing the `AssociatedBundleIdentifiers` key, required since macOS 13 for Background Task Management to associate the daemon with its app bundle.

A secondary usability issue: the server URL must be configured manually by editing `agent.cfg` after installation. The `sed`-based approach documented in guides is fragile because `agent.cfg` contains three commented `#server =` example lines with the same hostname prefix, causing all three to be activated at once.

### Fix: modern launchctl API (macOS 13+)

- **`contrib/macosx/scripts/postinstall`**
  - Add `AssociatedBundleIdentifiers: com.teclib.glpi-agent` to the generated plist
  - Replace `launchctl load` + `launchctl start` with `launchctl bootstrap system` on macOS 13+
  - Keep legacy commands for macOS 12 and earlier (backward compatible)

- **`contrib/macosx/scripts/preinstall`**
  - Replace `launchctl stop` + `launchctl unload` with `launchctl bootout system` on macOS 13+

- **`contrib/macosx/scripts/uninstaller.sh`**
  - Replace `launchctl stop` + `launchctl unload` with `launchctl bootout system` on macOS 13+

All three scripts use `$(sw_vers -productVersion | cut -d. -f1)` for runtime version detection and remain backward compatible with macOS 12 and earlier.

### Feat: server URL dialog during installation

- **`contrib/macosx/scripts/postinstall`**
  - Shows a native macOS dialog (`osascript`) asking for the GLPI server URL
  - Writes the result to `etc/conf.d/01-server.cfg` (picked up automatically via the existing `include "conf.d"` directive in `agent.cfg`)
  - Silently skipped when no console user is detected, preserving compatibility with unattended/MDM deployments

## Testing

Validated on **macOS 26 Tahoe (Apple Silicon M-series)**:
- `launchctl bootout` completes without error in `uninstaller.sh`
- `launchctl bootstrap system` starts the service correctly after installation
- Server URL dialog appears during installation and writes `conf.d/01-server.cfg`
- Agent connects to server on first run without manual `agent.cfg` edits

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


g-bougard:
Hi @CristianPizaTR 

please read this point in GLPI Contributing page: https://github.com/glpi-project/glpi/blob/11.0/bugfixes/CONTRIBUTING.md#ai-assisted-contributions

Here, your AI agent is set as co-authour in your PR. This point is still not acceptable.

Also it seems you asked it to produce the description. As developer, we won't discuss with an IA. So please, use vibe coding if you want, but behind that, handle by yourself the communication with developers.

To be honest, I take a quick review on the produced code, and I saw few things which are not correct or useless. Did you control by yourself the produced code ?

CristianPizaTR:
Hi, thanks for the feedback and for taking the time to review.

The problem I was trying to solve: on macOS 13 Ventura and later (tested on macOS 26 Tahoe, Apple Silicon), installing GLPI Agent results in the service never starting because launchctl load returns Load failed: 5: Input/output error. The same issue affects upgrades (preinstall) and uninstalls. The core fix is replacing load/unload with bootstrap/bootout with a runtime version check.

I'll close this PR and come back with a cleaner, focused contribution addressing the points you raised. I'll also make sure to review the code myself before submitting.

g-bougard:
Thank you.

Don't hesitate to split the PR in 2 or 3, making the others dependent on the first ones.

PR #1174, opened 2026-06-03
fix: use launchctl bootstrap/bootout on macOS 13 Ventura and later

CristianPizaTR:
Problem
On macOS 13 Ventura and later (tested on macOS 26 Tahoe, Apple Silicon), installing GLPI Agent leaves the service not running. The postinstall script calls launchctl load, which is deprecated and returns Load failed: 5: Input/output error on modern macOS. The same issue affects upgrades via preinstall and uninstalls via uninstaller.sh.

Fix
Replace launchctl load/unload/start with launchctl bootstrap/bootout on macOS 13+. A runtime check on sw_vers selects the right API, keeping backward compatibility with macOS 12 and earlier.

Tested on macOS 26 Tahoe (arm64). (AI-assisted development: Claude)

### DIFF
--- a/Changes
+++ b/Changes
@@ -21,6 +21,9 @@ toolbox:
 * Fix undefined error on Results page while tag_filter is not set
 
 packaging:
+* fix: Update macOS postinstall, preinstall and uninstaller to use launchctl
+  bootstrap/bootout on macOS 13 Ventura and later, where launchctl load/unload
+  is deprecated and returns an error
 * Update Windows packaging to use:
   - Perl 5.42.2
   - gcc 16.1.0posix-14.0.0-msvcrt-r2
--- a/contrib/macosx/scripts/postinstall
+++ b/contrib/macosx/scripts/postinstall
@@ -42,6 +42,7 @@ fi
 echo "Only Tiger or newer OS supported, using LaunchDaemons plists"
 TMPPLIST="$(touch com.teclib.glpi-agent.plist && echo com.teclib.glpi-agent.plist || mktemp -t com.teclib.glpi-agent)"
 TPATH="/Library/LaunchDaemons"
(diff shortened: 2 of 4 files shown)


g-bougard:
Hi @CristianPizaTR 

have you any official documentation link which explains launchctl load/unload is deprecated ?

CristianPizaTR:
The official documentation is the launchctl(1) man page shipped with macOS itself, which classifies load and unload under "LEGACY SUBCOMMANDS" and explicitly lists bootstrap / bootout as recommended alternatives. There is no separate Apple web page announcing this, the man page is Apple's authoritative reference for command-line tools.

The practical issue is what I observed on macOS 13+: launchctl load returns Load failed: 5: Input/output error for third-party LaunchDaemons, while launchctl bootstrap system works. If you know of a different root cause for that error, I'm open to hearing it, perhaps the real issue is something else entirely (binary signing, entitlements, etc.) and the fix should be different.

g-bougard (approved):
Hi @CristianPizaTR 

thank you for your answer.

Actually I only have x86_64 test vm running tahoe 26 and launchctl doesn't seem to report an error with these legacy subcommands. Maybe I'm missing something but I also can understand the problem only occurs on arm64.

Also the error "Load failed: 5: Input/output error" seems to tell the file is still loaded. Maybe this error only occurs in specific context.

I can see load/unload and start/stop subcommands are now legacy but the man page doesn't tell since which version they are considered as legacy. It also doesn't tell since which version bootstrap/bootout subcommands are supported. I only found [launchd.info](https://www.launchd.info/) as external source which seems to say bootstrap/bootout could be used since 10.10. I think we can assume using 13 is good to assume load/unload/start/stop are deprecated. But I really would like to know what Apple is saying on that point.

Anyway I'm okay to merge this change.

PR #1175, opened 2026-06-03
feat(remote_mgt): Add DWService module for remote management inventory

eduardomozart:
### Description
This PR introduces a new remote management module to detect [DWService](https://www.dwservice.net/) (a popular cross-platform remote desktop and management tool) and adds it to the `REMOTE_MGMT` inventory section, similar to the existing support for TeamViewer and AnyDesk.

### Technical Details & Implementation
Extracting information from DWService requires a specific approach because the agent does not store the web dashboard's "Friendly Name" in a standard plain-text configuration file. 

This module implements the following logic to safely and natively extract the data without relying on external Python wrappers or breaking active connections:

1. **Dynamic Path Detection:**
   - **Windows:** Interrogates the registry (`Uninstall\DWAgent` and `WOW6432Node`) for the `InstallLocation`, falling back to standard `Program Files` paths.
   - **macOS / Linux:** Uses the `GLPI::Agent::Tools::Unix::getProcesses()` API to dynamically inspect the active process list for the daemon (`DWAgentSvc.app`, `dwagsvc`). This safely accounts for custom user installation paths, and gracefully falls back to default directories like `/Library/DWAgent` or `/usr/share/dwagent`.

2. **Unique ID Extraction:**
   - Parses the local `config.json` file using `GLPI::Agent::Tools::getAllLines()` and `Cpanel::JSON::XS` to extract the unique agent `key`. This ensures a highly reliable, immutable `ID` for GLPI and avoids dependency issues with older JSON libraries on Windows builds.

3. **Friendly Name Interception (IPC Memory Map):**
   - The DWService background daemon passes the web dashboard "Friendly Name" to its local UI monitor via a custom, disk-backed memory-mapped file (`sharedmem/status_config.shm`).
   - This module uses a safe `eval` block to read this `.shm` binary file, unpacks its 4-byte big-endian length integer, decodes the custom JSON byte-offset header, and securely extracts the `name` payload via `substr()` (stripping the fixed-block space padding).
   - If the name is successfully intercepted, it is used as the ID / Display Name in GLPI. If the agent is offline or the memory file is temporarily unavailable, it gracefully falls back to using the unique `key` (ID).

### Inventory Output Example

```
C:\Program Files\GLPI-Agent> glpi-agent --force --logger=stderr --debug
...
[debug] Running GLPI::Agent::Task::Inventory::Generic::Remote_Mgmt::DWService
[debug] DWService: Active installation found at C:\Program Files\DWAgent
[debug] DWService: Preparing for inventory -> ID: xxXXxXXxxXxxxXXXxxXX, NAME: TI - ANONYMOUS-188
...
```

### Test Output

```
C:\Users\eduardo.oliveira\Documents\GitHub\glpi-agent>perl -Ilib -It/lib t\tasks\inventory\generic\remote_mgmt\dwservice.t
1..10
[debug] Logger backend Stderr initialized
ok 1 - DWService is enabled on Win32
[debug] DWService: Active installation found at C:\Program Files\DWAgent
[debug] DWService: Preparing for inventory -> ID: xxXXxXXxxXxxxXXXxxXX, NAME: TI - ANONYMOUS-188
ok 2 - One remote mgmt entry found on Win32
ok 3 - Correct DWService data extracted on Win32
ok 4 - DWService is enabled on Linux
[debug] DWService: Active installation found at /usr/share/dwagent
[debug] DWService: Preparing for inventory -> ID: xxXXxXXxxXxxxXXXxxXX, NAME: TI - ANONYMOUS-188
ok 5 - One remote mgmt entry found on Linux
ok 6 - Correct DWService data extracted on Linux
ok 7 - DWService is enabled on macOS
[debug] DWService: Active installation found at /Library/DWAgent
[debug] DWService: Preparing for inventory -> ID: xxXXxXXxxXxxxXXXxxXX, NAME: TI - ANONYMOUS-188
ok 8 - One remote mgmt entry found on macOS
ok 9 - Correct DWService data extracted on macOS
ok 10 - no warnings
```

image

image


lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/DWService.pm
```diff
@@ -0,0 +1,202 @@
...
+use warnings;
+
+use parent 'GLPI::Agent::Task::Inventory::Module';
+
+use GLPI::Agent::Tools;
+use Fcntl qw(SEEK_SET);
+use JSON::PP;
+
+our $runMeIfTheseChecksFailed = 1;
+
```
g-bougard: `$runMeIfTheseChecksFailed` must be an array ref of modules that have to fail to know if we can this one. As far I understand, this value is not required:
```suggestion
```

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/DWService.pm
```diff
@@ -0,0 +1,202 @@
...
+# --- Helper: Dynamically find installation paths ---
+sub _get_base_paths {
+    my @paths;
+    
+    if ($^O eq 'MSWin32') {
```
g-bougard: Don't use `$^O` unless you module can't be run with RemoteInventory. In place, use `GLPI::Agent::Tools` `OSNAME()` api:
```suggestion
    if (OSNAME eq 'MSWin32') {
```

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/DWService.pm
```diff
@@ -0,0 +1,202 @@
...
+        require GLPI::Agent::Tools::Win32;
```
g-bougard: Better include UNIVERSAL::require in module "use" head lines:
```
use UNIVERSAL::require;
```

And use this code as we still does in many module:
```suggestion
        GLPI::Agent::Tools::Win32->require();
```

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/DWService.pm
```diff
@@ -0,0 +1,202 @@
...
+        
+        # Check standard registry keys and WOW6432Node
+        foreach my $reg_key (
+            'HKEY_LOCAL_MACHINE/SOFTWARE/Microsoft/Windows/CurrentVersion/Uninstall/DWAgent',
+            'HKEY_LOCAL_MACHINE/SOFTWARE/WOW6432Node/Microsoft/Windows/CurrentVersion/Uninstall/DWAgent'
+        ) {
+            my $install_loc = GLPI::Agent::Tools::Win32::getRegistryValue(path => "$reg_key/InstallLocation");
+            $install_loc =~ s{[\\/]+$}{} if $install_loc;
+            push @paths, $install_loc if $install_loc && -d $install_loc;
```
g-bougard: As for `$^O`, the "-d" operator won't work in RemoteIvnentory case. Better use `GLPI::Agent::Tools` `has_folder()` api:
```suggestion
            push @paths, $install_loc if $install_loc && has_folder($install_loc);
```

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/DWService.pm
```diff
@@ -0,0 +1,202 @@
...
+        }
+        
+        # Windows fallbacks
+        push @paths, 'C:/Program Files/DWAgent', 'C:/Program Files (x86)/DWAgent';
+    } else {
+        # Dynamic process detection on Unix systems (Linux / macOS)
+        my $ps_cmd = $^O eq 'darwin' ? 'ps -A -o command' : 'ps -e -o args';
```
g-bougard: As before, don't use `$^O` but:
```suggestion
        my $ps_cmd = OSNAME eq 'darwin' ? 'ps -A -o command' : 'ps -e -o args';
```
But anyway, you should better use `getProcesses()` api from our `GLPI::Agent::Tools::Unix` module which also includes also a way to filter the list. Check still existing use cases in our code.

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/DWService.pm
```diff
@@ -0,0 +1,202 @@
...
+                if ($line =~ m{(/.*?)/native/DWAgentSvc\.app}i || 
+                    $line =~ m{(/.*?)/native/dwag(?:svc|ent)}i) {
+                    push @paths, $1;
+                }
+            }
+            close($ph);
+        }
+        
+        # macOS fallback
+        push @paths, '/Library/DWAgent' if $^O eq 'darwin';
```
g-bougard:
```suggestion
        push @paths, '/Library/DWAgent' if OSNAME eq 'darwin';
```

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/DWService.pm
```diff
@@ -0,0 +1,202 @@
...
+        
+        # Linux fallbacks
+        push @paths, '/usr/share/dwagent', '/opt/dwagent' if $^O eq 'linux';
```
g-bougard:
```suggestion
        push @paths, '/usr/share/dwagent', '/opt/dwagent' if OSNAME eq 'linux';
```

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/DWService.pm
```diff
@@ -0,0 +1,202 @@
...
+    }
+    
+    # Remove duplicates and ensure the directory exists
+    my %seen;
+    return grep { $_ && -d $_ && !$seen{$_}++ } @paths;
```
g-bougard:
```suggestion
    return grep { $_ && has_folder($_) && !$seen{$_}++ } @paths;
```

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/DWService.pm
```diff
@@ -0,0 +1,202 @@
...
+}
+
+sub isEnabled {
+    # Check if config.json exists in any of the found paths
+    foreach my $path (_get_base_paths()) {
+        return 1 if -f "$path/config.json";
```
g-bougard:
```suggestion
        return 1 if has_file("$path/config.json");
```

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/DWService.pm
```diff
@@ -0,0 +1,202 @@
...
+sub doInventory {
+    my (%params) = @_;
+    my $inventory = $params{inventory};
+    my $logger    = $params{logger};
+
+    my $base_path;
+
+    # 1. Locate the valid installation path
+    foreach my $path (_get_base_paths()) {
+        if (-f "$path/config.json") {
```
g-bougard:
```suggestion
        if (has_file("$path/config.json")) {
```

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/DWService.pm
```diff
@@ -0,0 +1,202 @@
+package GLPI::Agent::Task::Inventory::Generic::Remote_Mgmt::DWService;
+
+use strict;
+use warnings;
+
+use parent 'GLPI::Agent::Task::Inventory::Module';
+
+use GLPI::Agent::Tools;
+use Fcntl qw(SEEK_SET);
+use JSON::PP;
```
g-bougard: Don't use JSON::PP, it is no more provided in windows build.
```suggestion
use Cpanel::JSON::XS;
```

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/DWService.pm
```diff
@@ -0,0 +1,202 @@
...
+                # Matches absolute paths in memory, extracting the base directory
+                # macOS: /Library/DWAgent/native/DWAgentSvc.app/... -> /Library/DWAgent
+                # Linux: /usr/share/dwagent/native/dwagsvc -> /usr/share/dwagent
+                if ($line =~ m{(/.*?)/native/DWAgentSvc\.app}i || 
+                    $line =~ m{(/.*?)/native/dwag(?:svc|ent)}i) {
+                    push @paths, $1;
+                }
+            }
+            close($ph);
+        }
```
g-bougard: Adapt this part using getProcesses() as said earlier

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/DWService.pm
```diff
@@ -0,0 +1,202 @@
...
+
+    # 2. Extract the unique ID (key) from config.json
+    my $config;
+    eval {
+        local $/; 
+        open(my $fh, '<:encoding(UTF-8)', "$base_path/config.json") or die "Cannot open config: $!";
+        my $json_text = <$fh>;
+        close($fh);
+        $config = decode_json($json_text);
+    };
```
g-bougard, line 116: Use our getAllLines() api, it is compatible with RemoteInventory:
```suggestion
    my $json_text = getAllLines(
        file   => "$base_path/config.json",
        logger => $logger
    );
    return if empty($json_text);

    my $config;
    eval {
        $config = decode_json($json_text);
    };
```

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/DWService.pm
```diff
@@ -0,0 +1,202 @@
...
+        close($fh);
+    };
+
+    if ($@) {
+        $logger->debug("DWService: Failed to extract data from SHM - $@");
+        return;
+    }
+
+    return \%extracted_data;
+}
```
g-bougard, line 210: I won't enter in the details here, but this _extract_shm_data() function only works locally, use has_file() and getAllLines() api as still suggested and adapt your algorithm.

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/DWService.pm
```diff
@@ -0,0 +1,202 @@
...
+
+    $logger->debug("DWService: Preparing for inventory -> ID: $display_name, NAME: $display_name");
+    
+    # --- Commented out state and sessions_status extractions ---
+    # if ($shm_data && $shm_data->{'state'}) {
+    #      $logger->debug("DWService Extra Info: Current Agent state -> " . $shm_data->{'state'});
+    # }
+    # if ($shm_data && $shm_data->{'sessions_status'} && $shm_data->{'sessions_status'} ne '{}' && $shm_data->{'sessions_status'} ne '[]') {
+    #      $logger->debug("DWService Extra Info: The agent has active sessions! -> " . $shm_data->{'sessions_status'});
+    # }
```
g-bougard: Don't keep no more required commented code.
```suggestion
```

g-bougard (changes requested):
Hi @eduardomozart 

I'm okay to add this support, but your module needs to be fixed so it can be run during RemoteInventory.

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/DWService.pm
eduardomozart: Yes, you're right. I thought I already removed it. Thank you for pointing it out!

g-bougard:
Also, can you provide example files so I can add unittests ?

I would probably need:
* registry export of `HKEY_LOCAL_MACHINE/SOFTWARE/Microsoft/Windows/CurrentVersion/Uninstall/DWAgent` key
* a config.json sample
* and sharedmem/status_config.shm file sample

If you have different context, you can provide as much sample as possible.

If you can't provide test files, this will just mean we won't be able to check it still works as expected after an update.

eduardomozart (commented):
Hello @g-bougard, thank you for taking your time to review this PR!
I've updated this PR with the tests and test data, and also applied all the changes you've requested. Tests and agent inventory are reporting DWService data as expected. I also updated the PR description to match the changes applied.

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/DWService.pm
```diff
@@ -0,0 +1,205 @@
...
+    
+    # Eval block catches failures in case DWService alters the binary structure in the future
+    eval {
+        my $content = getAllLines(
+            file   => $shm_file,
+            mode   => '<:raw',
+            logger => $logger
+        );
+
+        die "Cannot read $shm_file" unless defined $content;
```
g-bougard: You should include `\n` at the end of the `die` message as we don't need the line where the error occurred. And you can use `empty()` api:
```suggestion
        die "Cannot read $shm_file\n" if empty($content);
```

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/DWService.pm
```diff
@@ -0,0 +1,205 @@
...
+
+        # Read the first 4 bytes (header length)
+        my $len_bytes = substr($content, 0, 4);
+        die "Could not read header length" unless length($len_bytes) == 4;
```
g-bougard:
```suggestion
        die "Could not read header length\n" unless length($len_bytes) == 4;
```

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/DWService.pm
```diff
@@ -0,0 +1,205 @@
...
+        
+        # Unpack as unsigned 32-bit Big-Endian integer
+        my $len_def = unpack("N", $len_bytes);
+
+        # Read the JSON header describing the byte offsets
+        my $json_header = substr($content, 4, $len_def);
+        die "Could not read JSON header" unless length($json_header) == $len_def;
```
g-bougard:
```suggestion
        die "Could not read JSON header\n" unless length($json_header) == $len_def;
```

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/DWService.pm
```diff
@@ -0,0 +1,205 @@
...
+        my $fields = decode_json($json_header);
+
+        # List of fields we want to extract from memory
+        # 'state' and 'sessions_status' kept in comments for future use
+        my @target_fields = ('name'); #, 'state', 'sessions_status');
+
+        foreach my $target (@target_fields) {
+            if (exists $fields->{$target}) {
+                my $data_pos  = $fields->{$target}->{'pos'};
+                my $data_size = $fields->{$target}->{'size'};
```
g-bougard: Just to be safe, guaranty data_pos and data_size are not empty and can be seen as a positive integer:
```suggestion
                next if empty($fields->{$target}->{'pos'}) || $fields->{$target}->{'pos'} !~ /^\d+$/;
                next if empty($fields->{$target}->{'size'}) || $fields->{$target}->{'size'} !~ /^\d+$/;
                my $data_pos  = int($fields->{$target}->{'pos'});
                my $data_size = int($fields->{$target}->{'size'});
```

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/DWService.pm
```diff
@@ -0,0 +1,205 @@
...
+
+                # Read the fixed block of bytes: 4 (length int) + JSON header length + data offset
+                my $raw_value = substr($content, 4 + $len_def + $data_pos, $data_size);
+
+                # DWService pads strings with spaces (" "), clear them with regex
+                $raw_value =~ s/\s+$// if defined $raw_value;
```
g-bougard: You can use `trimWhitespace()` here:
```suggestion
                # Read the fixed block of bytes: 4 (length int) + JSON header length + data offset
                # DWService pads strings with spaces (" "), clear them with regex
                my $raw_value = trimWhitespace(substr($content, 4 + $len_def + $data_pos, $data_size));
```

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/DWService.pm
```diff
@@ -0,0 +1,205 @@
...
+                
+                $extracted_data{$target} = $raw_value if defined $raw_value && $raw_value ne "";
```
g-bougard: Use `empty()` here:
```suggestion
                $extracted_data{$target} = $raw_value unless empty($raw_value);
```

t/tasks/inventory/generic/remote_mgmt/dwservice.t
```diff
@@ -0,0 +1,171 @@
...
+            return 1;
+        }
+        # Map Unix path
+        if ($file eq '/usr/share/dwagent/config.json') {
+            return 1;
+        }
+        if ($file eq '/usr/share/dwagent/sharedmem/status_config.shm') {
+            return 1;
+        }
+        return $original_has_file->($file);
```
g-bougard: Here you should better use `original` method and remove l.54:
```suggestion
        return $tools_module->original('has_file')->($file);
```

t/tasks/inventory/generic/remote_mgmt/dwservice.t
```diff
@@ -0,0 +1,171 @@
...
+    sub {
+        my (%params) = @_;
+        my $file = $params{file};
+        if ($file && ($file eq 'C:\Program Files\DWAgent/config.json' || $file eq 'C:\Program Files\DWAgent\config.json' || $file eq '/usr/share/dwagent/config.json')) {
+            $params{file} = "$test_base/config.json";
+        }
+        if ($file && ($file eq 'C:\Program Files\DWAgent/sharedmem/status_config.shm' || $file eq 'C:\Program Files\DWAgent\sharedmem\status_config.shm' || $file eq '/usr/share/dwagent/sharedmem/status_config.shm')) {
+            $params{file} = "$test_base/sharedmem/status_config.shm";
+        }
+        return $original_getAllLines->(%params);
```
g-bougard: Remove l.77 and use `original` api like this:
```suggestion
        return $tools_module->original('getAllLines')->(%params);
```

t/tasks/inventory/generic/remote_mgmt/dwservice.t
```diff
@@ -0,0 +1,171 @@
...
+# Mock has_file and has_folder to look in our test directory instead of the real C:/Program Files/DWAgent
+my $test_base = "t/resources/generic/dwservice";
+
+$tools_module->mock(
+    'has_folder',
+    sub {
+        my ($path) = @_;
+        return 1 if $path eq 'C:\Program Files\DWAgent';
+        return 1 if $path eq '/usr/share/dwagent';
+        return GLPI::Agent::Tools::has_folder($path);
```
g-bougard: You should use `original` api here:
```suggestion
        return $tools_module->original('has_folder')->($path);
```

t/resources/generic/dwservice/config.json
```diff
@@ -0,0 +1,15 @@
...
+ "key": "xxXXxXXxxXxxxXXXxxXX",
+ "listen_port": 7950,
+ "nodes": [
+  {
+   "id": "ND123456",
+   "port": "443",
+   "server": "node123456.dwservice.net"
+  }
+ ],
+ "password": "eJwLr8jyDfDx806rSA/x8C/wdIlwS/cqqcx3r6gqcI4CAKV6Cto=",
```
g-bougard: Can you confirm this is a forged password here ?

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/DWService.pm
```diff
@@ -0,0 +1,205 @@
...
+                $pf =~ s{\\}{/}g;
+                push @paths, "$pf/DWAgent";
+            }
+        }
+        # Hardcoded ultimate fallbacks
+        push @paths, 'C:/Program Files/DWAgent', 'C:/Program Files (x86)/DWAgent';
+    } else {
+        # Dynamic process detection on Unix systems (Linux / macOS)
+        GLPI::Agent::Tools::Unix->require();
+        my @processes = GLPI::Agent::Tools::Unix::getProcesses();
```
g-bougard: You should provide a filter here to optimize memory usage:
```suggestion
        my @processes = GLPI::Agent::Tools::Unix::getProcesses(
            filter    => qr{nativer/(dwag(?:ent|svc|entsvc\.app)}i,
            namespace => "same",
        );
```

g-bougard (changes requested):
I see few other things to update. Check my comments.

Also I see your shared a file which includes a password. Be sure the password is a fake.

t/resources/generic/dwservice/config.json
eduardomozart: It has been anonymized in https://github.com/eduardomozart/glpi-agent/commit/ffb8ffa1f50082a4e5e020ee2a1785a940351b1a commit to "ThisIsAnAnonymizedPassword1234".

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/DWService.pm
eduardomozart: There's a typo in "native". I fixed it in https://github.com/eduardomozart/glpi-agent/commit/ffb8ffa1f50082a4e5e020ee2a1785a940351b1a commit

eduardomozart (commented):
Done!

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/DWService.pm
```diff
@@ -0,0 +1,204 @@
+package GLPI::Agent::Task::Inventory::Generic::Remote_Mgmt::DWService;
+
+use strict;
+use warnings;
+
+use parent 'GLPI::Agent::Task::Inventory::Module';
+
+use GLPI::Agent::Tools;
+use Fcntl qw(SEEK_SET);
```
g-bougard: This is line is no more required as now you use GLPI::Agent dedicated api.
```suggestion
```

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/DWService.pm
```diff
@@ -0,0 +1,204 @@
...
+        GLPI::Agent::Tools::Win32->require();
+        
+        # Check standard registry keys and WOW6432Node
+        foreach my $reg_key (
+            'HKEY_LOCAL_MACHINE/SOFTWARE/Microsoft/Windows/CurrentVersion/Uninstall/DWAgent',
+            'HKEY_LOCAL_MACHINE/SOFTWARE/WOW6432Node/Microsoft/Windows/CurrentVersion/Uninstall/DWAgent'
+        ) {
+            my $install_loc = GLPI::Agent::Tools::Win32::getRegistryValue(path => "$reg_key/InstallLocation");
+            $install_loc =~ s{[\\/]+$}{} if $install_loc;
+            push @paths, $install_loc if $install_loc && has_folder($install_loc);
```
g-bougard: Little cleanup/optimization:
```suggestion
            my $install_loc = GLPI::Agent::Tools::Win32::getRegistryValue(path => "$reg_key/InstallLocation")
                or next;
            $install_loc =~ s{[\\/]+$}{};
            push @paths, $install_loc if has_folder($install_loc);
```

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/DWService.pm
```diff
@@ -0,0 +1,204 @@
...
+        }
+        
+        # Windows fallbacks using environment variables
+        foreach my $env (qw(ProgramFiles ProgramFiles(x86) ProgramW6432)) {
+            if (my $pf = $ENV{$env}) {
+                $pf =~ s{\\}{/}g;
+                push @paths, "$pf/DWAgent";
+            }
```
g-bougard: Your code is okay but this one use our more common structure:
```suggestion
            my $pf = $ENV{$env}
                or next;
            $pf =~ s{\\}{/}g;
            push @paths, "$pf/DWAgent";
```

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/DWService.pm
```diff
@@ -0,0 +1,204 @@
...
+        my @processes = GLPI::Agent::Tools::Unix::getProcesses(
+            filter    => qr{native/(dwag(?:ent|svc|entsvc\.app))}i,
+            namespace => "same"
+        );
+        
+        foreach my $process (@processes) {
+            # We use this regex solely to extract the base installation directory (e.g. /usr/share/dwagent)
+            if ($process->{CMD} =~ m{(/.*?)/native/}i) {
+                push @paths, $1;
+            }
```
g-bougard: To be honest, I really don't like this code. `_get_base_paths()` will be called at least one time on any unix computer and it means a `ps` command will be run for this check on them. This is really not optjimized.

Indeed do you mean DWService can be installed anywhere on the system and you can't find a simpler way for checking where it could have been installed ?

I'm not aware of how DWService can be installed so I have to trust you. But imho, if packages are provided by the editor, you should just check for the standard installation paths and that's all.

I see another problem but really not critical: anyone on the system can fake that inventory by creating a process matching the regex and create a config.json file so it will inject his DWService configuration. Of course this has no real interest, but this looks like a security hole anyway.

If you really want to keep this code part, you should includ a comment explaining why it should be kept.

g-bougard (changes requested):
Hi @eduardomozart 

I still see few required changes.

I want also you read carefully my comment on using `getProcesses()` and provide an answer on my questions there.

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/DWService.pm
```diff
@@ -0,0 +1,204 @@
...
+        $logger->debug("DWService: Failed to parse config.json - $@");
+        return;
+    }
+
+    my $dw_id = $config->{key} if $config;
+
+    if (!$dw_id) {
+        $logger->debug("DWService: Could not extract 'key' from config.json");
+        return;
+    }
```
g-bougard: I missed a problem here reported by AUTHOR tests: https://github.com/glpi-project/glpi-agent/actions/runs/27154685225/job/80285708117?pr=1175

```suggestion
    unless ($config && $config->{key}) {
        $logger->debug("DWService: Could not extract 'key' from config.json");
        return;
    }

    my $dw_id = $config->{key};
```

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/DWService.pm
```diff
@@ -0,0 +1,204 @@
...
+
+    # 3. Intercept local data from shared memory (SHM)
+    my $shm_data = _extract_shm_data("$base_path/sharedmem/status_config.shm", $logger);
+    
+    # Extract the friendly name (if available)
+    my $dw_name = $shm_data->{'name'} if $shm_data;
+    
+    # Fallback logic for Display Name: try friendly name, otherwise fall back to unique ID.
+    my $display_name = $dw_name ? $dw_name : $dw_id;
```
g-bougard: Also there's a problem here as `$dw_name` definition is conditional:
```suggestion
    # Fallback logic for Display Name: try extracted friendly name, otherwise fall back to unique ID.
    my $display_name = $shm_data && exists($shm_data->{'name'}) ? $shm_data->{'name'} : $dw_id;
```

g-bougard (changes requested):
After test launch, there's also some problem reported I missed. For details, see the AUTHOR tests reported here: https://github.com/glpi-project/glpi-agent/actions/runs/27154685225/job/80285708117?pr=1175

Here I didn't list all the lines with only spaces. Please clean them by yourself and take care of this problem in the future.

eduardomozart (commented):
Hello @g-bougard,
There's no problem, take your time to review this PR. I also want to make sure it's stable for end-users, and I'm glad for your code contributions as I handle more PHP code than Perl, but Perl has a language-specific approach, and I have some PHP vices that I'm glad you helped me to identify and optimize.
About DWService: It is difficult developing an approach that is Generic (independent) of the user's operating system without relying on `getProcesses` because the installer allows the user to select where to install DWService (custom path). This is a feature present on all operating systems.

Image

On macOS, we can rely on `/Library/LaunchDaemons` and `/Applications/DWService.app/` because it has a standard way to manage services and applications, but Linux operating systems can use systemd, init.d etc. and there's no installation package for us to detect the installation (like `dpkg`) like it does in Windows.
On Linux systems, we could rely on `/etc/dwagent` file to retrieve the installation path, which is created by the installer program. I would recommend checking the running processes to check if it's not a stale DWService installation, but I'm just fine if you decide to rely on it for security reasons.
The `dwagent.sh` shell script, which installs the DWService in Unix-like operating systems, can be found at: https://www.dwservice.net/pt_BR/download.html

I also contacted the DWService support team to get information about gathering the Friendly name information, which is the same approach I've used in Perl code:

```
De: DWService.net support <[support@dwservice.net](mailto:support@dwservice.net)>
Enviado: sexta-feira, 5 de junho de 2026 08:31
Para: Eduardo Oliveira
Assunto: [Externo]RE: Get friendly name of the agent
 
Hello Eduardo,
 
Thank you for contacting us with your question.
 
> It doesn't seems to exist an easy way to retrieve the DWService Agent name from local machine.
 
ℹ️ Actually, the only solution is the one you mentioned: retrieve it from the status_config.shm file. You can create a Python script to read the shared memory file (based on our code).
The Agent's code can be found here: [GitHub - dwservice/agent: DWService agent for Linux, Mac and Windows · GitHub](https://github.com/dwservice/agent)
 
We are available if you have any further questions.
Have a nice day.
 
Best regards,
László F.
Staff DWService
www.dwservice.net
```

I also analyzed the installer through Claude Sonnet; here are some automated findings that can help us decide which approach you consider better for the project. Please let me know your thoughts!

# DWAgent Detection Analysis (Linux & macOS)

The script is a self-extracting bash installer (`#!/bin/bash`) that embeds a tar archive starting at `__TARFILE_BEGIN__`. It extracts to `/tmp/dwagent_install<timestamp>/` and runs a Python installer. Below are all reliable ways to detect whether dwagent is installed and/or running.

---

## 1. Installation Directory

**Linux (default):**
```
/usr/share/dwagent/
```
Key files that confirm installation:
```
/usr/share/dwagent/agent.py
/usr/share/dwagent/versions.json
/usr/share/dwagent/native/dwagsvc        # service control script
/usr/share/dwagent/native/service.mng    # records which init system was used
/usr/share/dwagent/native/installer.ver
/usr/share/dwagent/dwagent.pid           # present while running
```

**macOS (default):**
```
/Library/DWAgent/
```

**Detection command:**
```sh
[ -d /usr/share/dwagent ] && echo "installed (Linux)" || echo "not found"
[ -d /Library/DWAgent ]   && echo "installed (macOS)" || echo "not found"
```

---

## 2. Configuration File

The installer writes a small JSON config to:
```
/etc/dwagent       # Linux — contains {"path": "/usr/share/dwagent"}
```

```sh
[ -f /etc/dwagent ] && echo "config found"
cat /etc/dwagent   # prints install path
```

---

## 3. Service — Linux (all 5 init systems)

The `install()` function in `dwagsvc` tries each init system in priority order and records the result in `<install_dir>/native/service.mng`. This file is the **most reliable single indicator** of which method was used.

### Priority order of `install()`

```
1. systemctl  (systemd available)
2. chkconfig  (RHEL/CentOS)
3. update-rc.d (Debian/Ubuntu)
4. rc-update  (OpenRC — Alpine/Gentoo)
5. rc.local   (/etc/rc.d/rc.local exists — Slackware/minimal)
```

### Primary artifacts per init system

| Init system | `service.mng` value | Primary file |
|---|---|---|
| **systemd** | `systemctl` | `/etc/systemd/system/dwagent.service` |
| **chkconfig** | `chkconfig` | `/etc/init.d/dwagent` (symlink → `dwagsvc`) |
| **update-rc.d** | `updatercd` | `/etc/init.d/dwagent` (symlink → `dwagsvc`) |
| **OpenRC** | `openrc` | `/etc/init.d/dwagent` (copy of `dwagent.openrc`) |
| **rc.local** | `rclocal` | `/etc/rc.d/rc.dwagent` (symlink → `dwagsvc`) |

### Secondary artifacts per init system

| Init system | Additional artifacts |
|---|---|
| **chkconfig** | `/etc/rc.d/rc[0-6].d/{K,S}??dwagent` (created by `chkconfig --add`) |
| **update-rc.d** | `/etc/rc?.d/S??dwagent` symlinks (created by `update-rc.d defaults`) |
| **OpenRC** | `/etc/runlevels/default/dwagent` (created by `rc-update add`) |
| **rc.local** | Entry in `/etc/rc.d/rc.local` and `/etc/rc.d/rc.local_shutdown` |

### Detection commands

```sh
# Most reliable: read service.mng directly
# Outputs one of: systemctl | chkconfig | updatercd | openrc | rclocal
cat /usr/share/dwagent/native/service.mng

# systemd
[ -f /etc/systemd/system/dwagent.service ] && echo "systemd unit found"
systemctl list-unit-files | grep dwagent

# chkconfig / update-rc.d / OpenRC — all share the same init.d path
[ -f /etc/init.d/dwagent ] && echo "init.d entry found"

# chkconfig — runlevel symlinks
ls /etc/rc.d/rc?.d/*dwagent 2>/dev/null

# update-rc.d — runlevel symlinks
ls /etc/rc?.d/*dwagent 2>/dev/null

# OpenRC — runlevel link
ls /etc/runlevels/*/dwagent 2>/dev/null

# rc.local — symlink + text entries
[ -f /etc/rc.d/rc.dwagent ] && echo "rc.local symlink found"
grep -q 'rc\.dwagent' /etc/rc.d/rc.local 2>/dev/null          && echo "rc.local entry found"
grep -q 'rc\.dwagent' /etc/rc.d/rc.local_shutdown 2>/dev/null && echo "rc.local_shutdown entry found"
```

---

## 4. Service — macOS (launchd)

The installer registers **three** launchd entries for the standard `DWAgent` name:

| Role | Label | Plist path |
|---|---|---|
| **Daemon** (system, root) | `net.dwservice.agsvc` | `/Library/LaunchDaemons/net.dwservice.agsvc.plist` |
| **Agent** (user session) | `net.dwservice.aglac` | `/Library/LaunchAgents/net.dwservice.aglac.plist` |
| **Systray** (user session) | `net.dwservice.agsystray` | `/Library/LaunchAgents/net.dwservice.agsystray.plist` |

Legacy paths also checked during uninstall (kept for compatibility):
```
/Library/LaunchDaemons/net.dwservice.agent.plist
/System/Library/LaunchDaemons/org.dwservice.agent.plist
```

**Detection commands:**
```sh
# Plist files (installed state)
ls /Library/LaunchDaemons/net.dwservice.*.plist 2>/dev/null
ls /Library/LaunchAgents/net.dwservice.*.plist 2>/dev/null

# Legacy paths
[ -f /Library/LaunchDaemons/net.dwservice.agent.plist ] && echo "legacy daemon plist found"
[ -f /System/Library/LaunchDaemons/org.dwservice.agent.plist ] && echo "legacy system plist found"

# Running state via launchctl
launchctl list | grep dwservice
launchctl list net.dwservice.agsvc 2>/dev/null
```

---

## 5. Application Bundle — macOS

`install_shortcuts()` copies the app bundle to:
```
/Applications/DWAgent.app
```

```sh
[ -d /Applications/DWAgent.app ] && echo "app bundle found"
```

---

## 6. Running Process

The agent runs as a Python process executing `agent.py` via the bundled runtime:

```sh
# Linux / macOS
ps aux | grep -E 'dwagsvc|agent\.py|dwagent' | grep -v grep

# Check PID file (Linux — present only while running)
cat /usr/share/dwagent/dwagent.pid 2>/dev/null

# macOS — query launchd for running state
launchctl list net.dwservice.agsvc 2>/dev/null
```

---

## 7. Symlinked Binaries (Linux shell integration)

`install_shortcuts()` → `install_bin()` creates symlinks in `/usr/bin` (or `/bin` as fallback):
```
/usr/bin/dwagent_monitor
/usr/bin/dwagent_configure
/usr/bin/dwagent_uninstall
```

```sh
command -v dwagent_monitor 2>/dev/null && echo "symlinks found"
ls /usr/bin/dwagent_* /bin/dwagent_* 2>/dev/null
```

---

## 8. Autostart Desktop Entry (Linux)

`install_auto_run_monitor()` copies the systray desktop file to:
```
/etc/xdg/autostart/dwagent_systray.desktop   # system-wide
~/.config/autostart/dwagent_systray.desktop   # per-user
```

```sh
[ -f /etc/xdg/autostart/dwagent_systray.desktop ] && echo "system autostart entry found"
[ -f ~/.config/autostart/dwagent_systray.desktop ] && echo "user autostart entry found"
```

---

## Summary Detection Script

```sh
#!/bin/sh
found=0

# 1. Install directories
[ -d /usr/share/dwagent ] && echo "[+] Install dir: /usr/share/dwagent" && found=1
[ -d /Library/DWAgent ]   && echo "[+] Install dir: /Library/DWAgent"   && found=1

# 2. Config file (Linux)
[ -f /etc/dwagent ] && echo "[+] Config file: /etc/dwagent" && found=1

# 3. Service manager marker (Linux)
SVC_MNG=$(cat /usr/share/dwagent/native/service.mng 2>/dev/null)
[ -n "$SVC_MNG" ] && echo "[+] service.mng: $SVC_MNG" && found=1

# 3a. systemd
[ -f /etc/systemd/system/dwagent.service ] && echo "[+] systemd unit found" && found=1

# 3b. SysV init.d (chkconfig / update-rc.d / OpenRC)
[ -f /etc/init.d/dwagent ] && echo "[+] init.d entry found" && found=1
ls /etc/rc.d/rc?.d/*dwagent 2>/dev/null && echo "[+] chkconfig runlevel links found"  && found=1
ls /etc/rc?.d/*dwagent      2>/dev/null && echo "[+] update-rc.d runlevel links found" && found=1
ls /etc/runlevels/*/dwagent 2>/dev/null && echo "[+] OpenRC runlevel link found"       && found=1

# 3c. rc.local
[ -f /etc/rc.d/rc.dwagent ] && echo "[+] rc.local symlink found" && found=1
grep -q 'rc\.dwagent' /etc/rc.d/rc.local          2>/dev/null && echo "[+] rc.local entry found"          && found=1
grep -q 'rc\.dwagent' /etc/rc.d/rc.local_shutdown 2>/dev/null && echo "[+] rc.local_shutdown entry found" && found=1

# 4. macOS launchd plists
ls /Library/LaunchDaemons/net.dwservice.*.plist 2>/dev/null && echo "[+] LaunchDaemon plist found" && found=1
ls /Library/LaunchAgents/net.dwservice.*.plist  2>/dev/null && echo "[+] LaunchAgent plist found"  && found=1
[ -f /Library/LaunchDaemons/net.dwservice.agent.plist ]          && echo "[+] Legacy daemon plist found"        && found=1
[ -f /System/Library/LaunchDaemons/org.dwservice.agent.plist ]   && echo "[+] Legacy system plist found"        && found=1

# 5. macOS app bundle
[ -d /Applications/DWAgent.app ] && echo "[+] /Applications/DWAgent.app found" && found=1

# 6. Running processes
pgrep -f 'dwagsvc|agent\.py' > /dev/null 2>&1 && echo "[+] dwagent process running" && found=1

# 7. Symlinked binaries (Linux)
command -v dwagent_monitor > /dev/null 2>&1 && echo "[+] Symlink dwagent_monitor found" && found=1

# 8. Autostart desktop entries (Linux)
[ -f /etc/xdg/autostart/dwagent_systray.desktop ] && echo "[+] System autostart entry found" && found=1
[ -f ~/.config/autostart/dwagent_systray.desktop ] && echo "[+] User autostart entry found"  && found=1

[ $found -eq 0 ] && echo "[-] dwagent not detected"
```

---

## Key Identifiers at a Glance

| Artifact | Value |
|---|---|
| Default install dir (Linux) | `/usr/share/dwagent/` |
| Default install dir (macOS) | `/Library/DWAgent/` |
| Config file (Linux) | `/etc/dwagent` |
| Init system marker | `<install_dir>/native/service.mng` → `systemctl\|chkconfig\|updatercd\|openrc\|rclocal` |
| systemd unit | `/etc/systemd/system/dwagent.service` |
| SysV/OpenRC init | `/etc/init.d/dwagent` |
| chkconfig runlevel links | `/etc/rc.d/rc?.d/*dwagent` |
| update-rc.d runlevel links | `/etc/rc?.d/*dwagent` |
| OpenRC runlevel link | `/etc/runlevels/default/dwagent` |
| rc.local symlink | `/etc/rc.d/rc.dwagent` |
| rc.local text entry | `/etc/rc.d/rc.local` (grep `rc.dwagent`) |
| macOS LaunchDaemon label | `net.dwservice.agsvc` |
| macOS LaunchAgent label | `net.dwservice.aglac` |
| macOS Systray label | `net.dwservice.agsystray` |
| macOS app bundle | `/Applications/DWAgent.app` |
| PID file (while running) | `<install_dir>/dwagent.pid` |
| Process command match | `dwagsvc`, `agent.py` |

g-bougard:
If there's a feature to install it on custom path, isn't there a way to detect the path without looking for a running process ? What happen if you run the DWService installer and it is still installed ? My guess is it must have some algorithm to detect installation. This is important when you need to upgrade. And I don't think it is based on running processes.

Actually, it seems you can analyze `/Library/LaunchDaemons/net.dwservice.agsvc.plist` on MacOS. It should give your the installation path.
And for linux, it seems you should probably just have to check all "Primary artifacts" and then you can even check for `<install_dir>/dwagent.pid` to check if it is really running.

This is more work, but this should make your module stronger.

eduardomozart (commented):
I've installed DWService on Linux (Ubuntu Server 24.04), macOS Tahoe and Windows 11, updated the test mocks to reproduce these environments and it's detecting DWService as expected without `getProcesses()`.

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/DWService.pm
```diff
@@ -0,0 +1,215 @@
...
+            push @paths, "$pf/DWAgent";
+        }
+        # Hardcoded ultimate fallbacks
+        push @paths, 'C:/Program Files/DWAgent', 'C:/Program Files (x86)/DWAgent';
+    } elsif (OSNAME eq 'darwin') {
+        # macOS: extract install path from LaunchDaemon plist
+        my $plist = '/Library/LaunchDaemons/net.dwservice.agsvc.plist';
+        if (has_file($plist)) {
+            my $path = _get_path_from_plist($plist);
+            push @paths, $path if $path;
```
g-bougard: Instead of using a regex on the file, we have a helper for xml plist with GLPI::Agent::XML library. Can you try the following ?
```suggestion
            GLPI::Agent::XML->require();
            my $xml = GLPI::Agent::XML->new(
                file     => $plist,
                is_plist => 1,
            );
            if ($xml) {
                my $dump = $xml->dump_as_hash();
                my $path = $dump && $dump->{plist} && $dump->{plist}->{ProgramArguments} ? $dump->{plist}->{ProgramArguments}->[1] : "";
                push @paths, $path if $path;
            }
```

Here is a one-liner to show you how GLPI::Agent::XML works with plist xml files:
```shell
[glpi-agent]$ perl -Ilib -MGLPI::Agent::XML -MData::Dumper -e '$p = GLPI::Agent::XML->new(file => "/tmp/net.dwservice.agsvc.plist", is_plist => 1); $x = $p->dump_as_hash(); print Dumper($x); print "PATH: $x->{plist}->{ProgramArguments}->[1]\n"'
$VAR1 = {
          'plist' => {
                       'RunAtLoad' => undef,
                       'Label' => 'net.dwservice.agsvc',
                       'StandardErrorPath' => '/Library/DWAgent/native/service.log',
                       'KeepAlive' => undef,
                       'OnDemand' => undef,
                       'ExitTimeOut' => '60',
                       'StandardOutPath' => '/Library/DWAgent/native/service.log',
                       'ProgramArguments' => [
                                               '/Library/DWAgent/native/DWAgentSvc.app/Contents/MacOS/DWAgentSvc',
                                               '/Library/DWAgent',
                                               'run'
                                             ]
                     }
        };
PATH: /Library/DWAgent
```

You can also secure and simplify the running using an eval and ignoring checks:
```suggestion
            eval {
                GLPI::Agent::XML->require();
                my $xml = GLPI::Agent::XML->new(
                    file     => $plist,
                    is_plist => 1,
                )->dump_as_hash();
                my $path = $xml->{plist}->{ProgramArguments}->[1];
                push @paths, $path if $path;
            };
```

If you prefer, you can also better use this code into your _get_path_from_plist() function.

lib/GLPI/Agent/Task/Inventory/Generic/Remote_Mgmt/DWService.pm
```diff
@@ -0,0 +1,215 @@
+package GLPI::Agent::Task::Inventory::Generic::Remote_Mgmt::DWService;
+
+use strict;
+use warnings;
+
+use parent 'GLPI::Agent::Task::Inventory::Module';
+
+use GLPI::Agent::Tools;
+use Cpanel::JSON::XS;
```
g-bougard: I just realize you're missing a `use` here as we use `require()` api syntax. I also prefer we call `use` on system libraries before GLPI::Agent ones:
```suggestion
use UNIVERSAL::require;
use Cpanel::JSON::XS;

use GLPI::Agent::Tools;
```

g-bougard (changes requested):
Hi @eduardomozart 

thank you for your effort, I appreciate.

I saw a `use` is missing and I've a suggestion to use our dedicated XML library to directly read and parse the plist file.

g-bougard:
Also AUTHOR test is still failing on lines with only spaces: https://github.com/glpi-project/glpi-agent/actions/runs/27212607449/job/80534509306?pr=1175

Please, fix that problem.

It seems the plist test file you added is also triggering `t/07whitespaces.t`. In that case, I will accept you add the following patch to your PR:
```diff
diff --git a/t/07whitespaces.t b/t/07whitespaces.t
index 4ae63b52b..1fc19d37f 100755
--- a/t/07whitespaces.t
+++ b/t/07whitespaces.t
@@ -14,5 +14,5 @@ plan(skip_all => 'Test::Whitespaces required')
 
 Test::Whitespaces->use({
     dirs   => [ qw/lib bin t/],
-    ignore => [ qr/~$/, qr/mock.t$/, qr/cisco.t$/ ],
+    ignore => [ qr/~$/, qr/mock\.t$/, qr/cisco\.t$/, qr/\.plist$/ ],
 });
```

eduardomozart (commented):
I merged your suggestions, and both tests seem to be passing.

g-bougard (changes requested):
There still a problem with AUTHOR tests: https://github.com/glpi-project/glpi-agent/actions/runs/27274206510/job/80555651210?pr=1175
1. You still didn't remove lines with only spaces. This is the third time I ask you this on this PR. Please read all my previous comments.
2. You forgot to remove your `_get_path_from_plist()` internal api
3. During test, we can see debug output: https://github.com/glpi-project/glpi-agent/actions/runs/27274206510/job/80555651210?pr=1175#step:7:123

So please, [run AUTHOR test as GH Actions does](https://github.com/glpi-project/glpi-agent/blob/d56575ffb1e4ddf4a48b371ae4ebb73f6ec62589/.github/workflows/glpi-agent-ci.yml#L53) before requesting a review on your PRs:
```shell
make test TEST_AUTHOR=1 TEST_FILES="t/02critic.t t/03var.t t/04pod-syntax.t t/06pod-spelling.t t/06pod-listed-categories.t t/07whitespaces.t t/09cpanmeta.t"
```
If you can't run it locally, enable GH Actions on your fork repo and wait ci-workflow is okay before submitting.

eduardomozart (commented):
I've applied your tests; I didn't know there was a Q&A test, so I added those tests into my suite. For next PRs, I'll run them.
The GitHub Actions doesn't seem to be triggered by me, only when a maintainer requests it. I think it could work like GLPI: every PR always runs them, and the maintainers only review your PR if all tests are passing.

t/tasks/inventory/generic/remote_mgmt/dwservice.t
```diff
@@ -0,0 +1,204 @@
...
+    sub { return $mock_osname; }
+);
+
+# Use a standard logger with Stderr backend so debug messages don't crash the test
+my $logger = GLPI::Agent::Logger->new(
+    config => {
+        debug  => 2,
+        logger => ['Stderr']
+    }
+);
```
g-bougard: For tests, and to avoid output anything not useful, you can use a test dedicated logger:
```suggestion
# Use a test logger so debug messages don't crash the test
my $logger = GLPI::Agent::Logger->new(
    logger => [ 'Test' ]
);
```

g-bougard (changes requested):
For debug output during tests, you indeed have to change the used logger configuration.

Can you test my suggestion ?

Edit: Your added `if $logger` on related lines in the previous commit is not sufficient.

t/tasks/inventory/generic/remote_mgmt/dwservice.t
```diff
@@ -0,0 +1,204 @@
...
+        return loadRegistryDump("t/resources/generic/dwservice/dwservice-win32-uninstall.reg");
+    }
+);
+
+# Mock tools
+my $tools_module = Test::MockModule->new(
+    'GLPI::Agent::Task::Inventory::Generic::Remote_Mgmt::DWService'
+);
+# Mock has_file and has_folder to look in our test directory instead of the real C:/Program Files/DWAgent
+my $test_base = "t/resources/generic/dwservice";
```
g-bougard: I just realized this path is wrong: Actually all our test resources are under `resources` base folder.

This is also the reason plist file was reported by `t/07-whitespaces.t` test.

So you can revert the `t/07-whitespaces.t` patch and move all your test files under `resources/generic/dwservice` and then update your test code accordingly.

t/07whitespaces.t
```diff
@@ -14,5 +14,5 @@ plan(skip_all => 'Test::Whitespaces required')
 
 Test::Whitespaces->use({
     dirs   => [ qw/lib bin t/],
-    ignore => [ qr/~$/, qr/mock.t$/, qr/cisco.t$/ ],
+    ignore => [ qr/~$/, qr/mock\.t$/, qr/cisco\.t$/, qr/\.plist$/ ],
```
g-bougard: Finally you won't need this patch if your resources test files are in the right place. So you can revert it.

g-bougard (changes requested):
Just noticed your resources test files was in the wrong place. See my comments.

eduardomozart (commented):
I moved the resources to the correct folder and updated the `logger` test. Tests and Q&A passing.

g-bougard (approved):
Thank you for your efforts.

PR #1176, opened 2026-06-03
feat(win32): collect network adapter description, manufacturer, model, mtu, virtualdev and link status from WMI

eduardomozart:
## Summary

On Windows, the hardware inventory for network cards was only showing the MAC address in GLPI's component list, without any network card name. This happened because the `MANUFACTURER` and `MODEL` fields were never populated for the `NETWORKS` section. 

Additionally, GLPI users expect the network adapter `DESCRIPTION` field to display the Windows connection name (e.g., "Ethernet" or "Wi-Fi" from `ncpa.cpl`). This provides parity with Unix systems, where this field typically holds the logical interface name (like `eth0` or `wlan0`).

Finally, setups using NIC Teaming or LoadBalance multiplexors had two major reporting inaccuracies:
1. Physical network adapters belonging to the team were incorrectly reported with a `STATUS` of "Down" even when physically connected, because the agent relied on IP address assignment to determine link status.
2. The virtual teamed adapter (e.g., "LoadBalance") was erroneously reported as a physical hardware device (`virtualdev: false`).

## Root Cause

1. The `getInterfaces()` function in `Tools/Win32.pm` queried `MSFT_NetAdapter` (Win8+) and `Win32_NetworkAdapter` (legacy) but did not request the driver/manufacturer, connection name, physical connection state, hardware MTU, or virtual status properties.
2. `NetAdapter.pm` had no data to populate `MANUFACTURER` and `MODEL` in the base interface, and passed the raw hardware description to the `DESCRIPTION` field instead of the connection name (e.g., "Ethernet" or "Wi-Fi" from `ncpa.cpl`).
3. `NetAdapter.pm` historically derived the `STATUS` of an interface by checking `IPEnabled` in `Win32_NetworkAdapterConfiguration`. In LoadBalance scenarios, the underlying physical NICs do not have IP addresses directly assigned, causing them to falsely report as "Down".
4. The `_isVirtual()` method relied on string matching (e.g., `ROOT\` or "RAS Adapter") to identify virtual interfaces, completely missing modern software-defined network adapters like the `COMPOSITEBUS` multiplexor used in LoadBalancing.
5. The agent relied solely on `Win32_NetworkAdapterConfiguration` for MTU, which is often empty or unreliable on modern Windows environments unless explicitly set in the registry.

### `lib/GLPI/Agent/Tools/Win32.pm`
- Added `DriverProvider`, `DriverDescription`, `Name`, `MediaConnectState`, `Virtual`, and `MtuSize` to the `MSFT_NetAdapter` WMI query (Windows 8 and above).
- Added `Manufacturer`, `Name`, `NetConnectionID`, and `NetConnectionStatus` to the `Win32_NetworkAdapter` query (Windows XP/7 legacy fallback).

### `lib/GLPI/Agent/Tools/Win32/NetAdapter.pm`
- Added `MANUFACTURER` and `MODEL` fields to `getBaseInterface()`.
- Added `_getManufacturer()`: returns `DriverProvider` (MSFT) or `Manufacturer` (legacy).
- Added `_getModel()`: returns `DriverDescription` (MSFT) or `Name` (legacy).
- Added `_getConnectionName()`: returns the connection name (`Name` for MSFT, `NetConnectionID` for legacy).
- Added `_getMtu()`: extracts the MTU natively from `MtuSize` (MSFT) if available, falling back to `Win32_NetworkAdapterConfiguration` (legacy).
- Extracted original description fallback logic into `_getHardwareDescription()` to safely preserve legacy VPN/RAS adapter validation in `_isVirtual()` and `getInterfaces()`.
- Updated `_isVirtual()` to directly query the native `Virtual` WMI property if available, accurately flagging LoadBalance and other modern software adapters as `virtualdev`.
- Updated `_getDescription()` to prioritize the connection name (e.g., "Ethernet" or "Wi-Fi" from `ncpa.cpl`) for GLPI reporting, making it equivalent to Unix logical interface names (e.g. `eth0`).
- Added `_getStatus()` to check the adapter's physical link state via `MediaConnectState` or `NetConnectionStatus` instead of relying solely on IP configuration, fixing the "Down" status for physical NICs in Teaming/LoadBalance setups.

### Tests
- Updated expectations in `t/agent/tools/win32.t` to accommodate the natively extracted MSFT WMI properties and modern `MtuSize` values.

## Field Mapping

| GLPI Field | Win8+ (`MSFT_NetAdapter`) | Legacy (`Win32_NetworkAdapter`) | Post-Processing |
|---|---|---|---|
| `MANUFACTURER` | `DriverProvider` | `Manufacturer` | None |
| `MODEL` | `DriverDescription` | `Name` | None |
| `DESCRIPTION` | `Name` | `NetConnectionID` | None |
| `MTU` | `MtuSize` | `Win32_NetworkAdapterConfiguration` | None |
| `STATUS` | `MediaConnectState` | `NetConnectionStatus` | Fallback to `IPEnabled` |
| `VIRTUALDEV` | `Virtual` | Legacy string matching | None |


g-bougard:
Hi @eduardomozart 

thank you for your PR, but I see one big problem: you're attempting to change the inventory format in GLPI::Agent::Inventory. And indeed, the final format doesn't match the specified one: https://github.com/glpi-project/inventory_format

So actually, we can't accept your PR.

If this actually works with GLPI, this may be a coincidence but this is finally not correct.

eduardomozart:
Hello,
It should match the GLPI inventory data; there's an open PR that uses this new data on GLPI at: https://github.com/glpi-project/glpi/pull/24471
There is, in fact, a limitation in GLPI regarding the stats information. I think transmitting it would require checking the GLPI version, because actually there is an artificial limitation in GLPI regarding the stats for non-NetworkEquipment objects (like Computer assets). I think I'll split the stats on a new PR https://github.com/glpi-project/glpi-agent/pull/1184 until https://github.com/glpi-project/glpi/pull/24471 is approved

g-bougard:
Hi Eduardo,

tests are failing actually. Can you fix them ?

eduardomozart:
Yes, it's now fixed. The firewall test was still using the old hardware description. Now it's passing:
```
1..5
ok 1 - test windows 10 FirewallPolicy: extract firewall profiles from registry
ok 2 - test windows 10 _getFirewallProfiles()
ok 3 - test windows 7_firewall FirewallPolicy: extract firewall profiles from registry
ok 4 - test windows 7_firewall _getFirewallProfiles()
ok 5 - no warnings
```

lib/GLPI/Agent/Tools/Win32/NetAdapter.pm
```diff
@@ -154,10 +195,48 @@ sub _getPNPDeviceID {
...
+    return $self->_getHardwareDescription();
+}
+
+sub _getStatus {
+    my ($self) = @_;
+
+    # MSFT_NetAdapter uses MediaConnectState: 1 = Connected, 2 = Disconnected
+    if (defined $self->{MediaConnectState}) {
+        return $self->{MediaConnectState} == 1 ? 'Up' : 'Down';
+    }
```
g-bougard: You should validate MediaConnectState is a number:
```suggestion
    if (defined($self->{MediaConnectState}) && $self->{MediaConnectState} =~ /^\d+$/) {
        return int($self->{MediaConnectState}) == 1 ? 'Up' : 'Down';
    }
```

lib/GLPI/Agent/Tools/Win32/NetAdapter.pm
```diff
@@ -154,10 +195,48 @@ sub _getPNPDeviceID {
...
+
+    # Win32_NetworkAdapter uses NetConnectionStatus: 2 = Connected, 7 = Media Disconnected
+    if (defined $self->{NetConnectionStatus}) {
+        return $self->{NetConnectionStatus} == 2 ? 'Up' : 'Down';
+    }
```
g-bougard: You should validate NetConnectionStatus is a number;
```suggestion
    if (defined($self->{NetConnectionStatus}) && $self->{NetConnectionStatus} =~ /^\d+$/) {
        return int($self->{NetConnectionStatus}) == 2 ? 'Up' : 'Down';
    }
```

g-bougard (changes requested):
Hi @eduardomozart 

I see a little syntax safety update is required as you test a numeric result.

I'm okay for the rest.

eduardomozart (commented):
Hello @g-bougard, it's great to see you! I agree and applied your changes as requested.

lib/GLPI/Agent/Tools/Win32/NetAdapter.pm
```diff
@@ -140,8 +145,44 @@ sub _getObjectIndex {
...
+sub _getMtu {
+    my ($self) = @_;
+
+    # MSFT_NetAdapter (Win8+) exposes MtuSize
+    return $self->{MtuSize} if $self->{MtuSize};
+
+    # Fallback to Win32_NetworkAdapterConfiguration MTU which might be unset
+    return $self->{_config}->{MTU} if $self->{_config}->{MTU};
+
+    return undef;
```
g-bougard: AUTHOR test is failing here, just remove `undef`:
```suggestion
    return;
```

g-bougard (changes requested):
AUTHOR is failing, please fix it.

eduardomozart (commented):
Done!

lib/GLPI/Agent/Tools/Win32/NetAdapter.pm
```diff
@@ -40,12 +40,14 @@ sub getBaseInterface {
         PNPDEVICEID => $self->_getPNPDeviceID(),
         MACADDR     => $self->{_config}->{MACADDR},
         DESCRIPTION => $self->_getDescription(),
-        STATUS      => $self->{_config}->{STATUS},
-        MTU         => $self->{_config}->{MTU},
+        STATUS      => $self->_getStatus(),
+        MTU         => $self->_getMtu(),
```
g-bougard: After the last update, _getMtu() can return no data and then we can have a odd number of elements in anonymous hash.

You should probably remove it from the anonymous hash and does like MODEL and MANUFACTURER to add it.

This way, you may also have to remove any test line where MTU is undef.

Please run and validate tests before committing.

g-bougard (changes requested):
Test are broken after last update. Check my commen to fix it.

lib/GLPI/Agent/Tools/Win32/NetAdapter.pm
eduardomozart: I've pushed a fix that removes `MTU` from the anonymous hash entirely, appending it dynamically afterwards just like we do for `MODEL` and `MANUFACTURER`. I've also stripped the `MTU => undef` assertions from `win32.t`.

eduardomozart (commented):
The test suite now passes locally with zero warnings.

g-bougard (approved):

PR #1178, opened 2026-06-05
feat: Accept Windows '/?' as --help in scripts

eduardomozart:
**Summary**
Windows users typically use `/?` to pull up command-line help. Currently, `Getopt::Long` ignores `/?` since it doesn't use standard dash prefixes. This causes scripts like `glpi-agent.bat` and `glpi-inventory.bat` to launch normally instead of showing the help text.

This PR adds a one-line mapping to map `/?` to `--help` inside the `@ARGV` array for all CLI scripts in the `bin/` directory. Treat the Windows standard help argument '/?' as '--help' in CLI scripts so GetOptions recognizes it. No other behavior changes were made.

**Changes**
- Added `map { $_ = '--help' if $_ eq '/?' } @ARGV;` to the arguments pipeline in all `bin/glpi-*` executables.
- Ensures consistency across the following executables: bin/glpi-agent, bin/glpi-esx, bin/glpi-injector, bin/glpi-inventory, bin/glpi-netdiscovery, bin/glpi-netinventory, bin/glpi-remote, bin/glpi-wakeonlan, and bin/glpi-win32-service.

### DIFF
--- a/bin/glpi-agent
+++ b/bin/glpi-agent
@@ -18,6 +18,9 @@ Getopt::Long::Configure( "no_ignorecase" );
 
 my $options = {};
 
+# Support Windows standard help argument /?
+map { $_ = '--help' if $_ eq '/?' } @ARGV;
+
 unless (GetOptions(
     $options,
     'assetname-support=i',
--- a/bin/glpi-esx
+++ b/bin/glpi-esx
@@ -21,6 +21,9 @@ my $options = {
     path    => '-'
 };
 
+# Support Windows standard help argument /?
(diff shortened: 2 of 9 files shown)


g-bougard:
Hi @eduardomozart 

But sorry, this feature won't be merged as it has no real interest. An option exists to output help, `/?` has always been specific to windows commands. Our scripts are not specific to windows. IMHO, if windows people starts to see such support they then will claim to support `/` for options.

So no, supporting `/?` just won't happen.

PR #1179, opened 2026-06-05
feat: Validate `no-category` parameter in Inventory task

eduardomozart:
## What does this PR do?
This PR fixes an issue where providing an invalid or unsupported category to the `--no-category` parameter (e.g., typing `processes` instead of `process`) would be silently ignored by the agent. 

## Why is this important?
Previously, an invalid category would result in the agent continuing the inventory run and collecting data that the user intended to exclude. This could lead to unintended data collection and privacy concerns. 

## How does it work?
Validation logic was added to `GLPI::Agent::Task::Inventory::run()` to check the user-provided `--no-category` values against the known supported categories returned by `getCategories()`. If an unknown category is detected, the agent will now log an error (e.g., `Unknown category 'processes' in no-category parameter`) and safely abort the inventory run if any invalid categories are found.

### DIFF
--- a/lib/GLPI/Agent/Task/Inventory.pm
+++ b/lib/GLPI/Agent/Task/Inventory.pm
@@ -158,14 +158,11 @@ sub run {
     # Validate no-category parameter against supported categories
     if (@{$self->{config}->{'no-category'}}) {
         my @categories = $self->getCategories();
-        my $invalid_category = 0;
         foreach my $category (@{$self->{config}->{'no-category'}}) {
             unless (any { $_ eq $category } @categories) {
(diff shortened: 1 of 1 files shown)


g-bougard (changes requested):
Hi @eduardomozart 

I'm okay to add an error message in log, but I won't accept to change the default to abort inventory on wrong setup category.

Actually it would mean glpi-agent could become no more working after an upgrade if a wrong category was kept in no-category params. People won't accept their agent to fail event if this is for a good reason.

So remove the code aborting the run.

eduardomozart (commented):
Done!

PR #1181, opened 2026-06-06
fix(videos): fallback to `pci.ids` database for generic video drivers on Windows

eduardomozart:
### Problem
When Windows relies on a generic fallback video driver (such as the "Microsoft Basic Display Adapter" or "Standard VGA Graphics Adapter"), the hardware values reported by WMI (`Win32_VideoController`) are often generic, empty, or severely truncated. For instance, `VideoProcessor` is typically cut off at exactly 31 characters (e.g., `Intel(R) BXT Mobile/Desktop Gra`), resulting in incomplete `<VIDEOS>` inventory records.

### Solution
This PR introduces a robust fallback mechanism in `GLPI::Agent::Task::Inventory::Win32::Videos` to accurately identify video controllers even when generic drivers are in use:

1. **Robust Generic Driver Detection:** Instead of relying on fragile string matching, the agent now iterates through the Windows Registry (`HKLM\SYSTEM\CurrentControlSet\Control\Class\{4d36e968-e325-11ce-bfc1-08002be10318}`). If no registry key matching the exact Plug and Play ID is found (because generic drivers use the generic class-code `PCI\CC_0300` as their `MatchingDeviceId`), the agent actively identifies that a generic driver is in control.
2. **PCI Database Translation:** Upon detecting a generic driver, the agent extracts the `Vendor ID`, `Device ID`, and `SubVendor ID` from the WMI `PNPDeviceID`. It then queries the local `share/pci.ids` database via `getPCIDeviceVendor()`.
3. **Logic Aligned with Linux:** The parsing logic mirrors the Linux counterpart (`Generic::PCI::Videos`), successfully splitting the `Family [Model]` syntax into `NAME` and `CHIPSET`, and gracefully handling `SubVendor` string prepending.
4. **Performance:** The heavy operation of reading the `pci.ids` file is strictly isolated to the fallback block. It will never run on machines that already have proper vendor drivers installed, preserving current behavior and maximum performance.

### Testing
- Added a new mock test case `microsoft-basic-display` in `t/tasks/inventory/windows/video.t`.
- Included mocked WMI and Registry dumps simulating the exact behavior of the generic Microsoft driver, successfully validating the `NAME` and `CHIPSET` overrides against the mock PCI database.

PR #1183, opened 2026-06-08
feat(win32): Gather PCISLOT for Videos and Controllers inventory

eduardomozart:
### Description
This PR introduces the retrieval of the `PCISLOT` property for both the `VIDEOS` and `CONTROLLERS` inventory sections on Windows environments.

To determine the accurate PCI address, the agent now looks up the device's `PNPDeviceID` within the `HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Enum` registry tree and parses the `LocationInformation` value. It effectively extracts the PCI `(bus, device, function)` mapping and formats it accurately (e.g., `00:02.0`) for GLPI.

### Changes Made
* **`Win32::Videos`**: Added logic to resolve and format `PCISLOT` during `_getVideos` iteration.
* **`Win32::Controllers`**: Replicated the `PCISLOT` discovery mechanism before parsing out `deviceid`.
* **Tests**: 
   * Created a new registry mock fixture (`microsoft-basic-display-3&11583659&0&10.reg`) containing `LocationInformation` data.
   * Updated `t/tasks/inventory/windows/video.t` to assert that `PCISLOT => '00:02.0'` is properly gathered.
   * Added a brand new test suite `t/tasks/inventory/windows/controllers.t` for `Win32::Controllers` to secure coverage and prevent regressions, utilizing the existing `microsoft-basic-display` WMI dump.

### Motivation
Providing accurate `PCISLOT` properties helps GLPI correctly map, merge, and identify physical hardware components on Windows identical to the behavior provided on Linux hosts via `lspci`. 

### Testing
Ran local test suites (`video.t` and `controllers.t`) simulating Windows WMI output and registry values, resulting in successful validation.


t/tasks/inventory/windows/controllers.t
```diff
@@ -0,0 +1,83 @@
...
+BEGIN {
+    # use mock modules for non-available ones
+    push @INC, "t/lib/fake/windows" if $OSNAME ne "MSWin32";
+}
+
+use Config;
+# check thread support availability
+if (!$Config{usethreads} || $Config{usethreads} ne "define") {
+    plan skip_all => "thread support required";
+}
```
g-bougard: Are you sure you need this check ? Actually, I don't this requirement in your code. As I know you're using AI agents, I can only suspect you just copied this test file from another one as template and didn't check if everything is really required.

Can you test without that check so ? And finally remove it if not required.

eduardomozart (commented):
Done.

g-bougard:
It seems tests are broken. Please fix it.

eduardomozart (commented):
It should pass the tests now.

g-bougard (changes requested):
Indeed it doesn't.

First it seems there's some conflict. You should first rebase your branch against current upstream develop. Maybe one of your previous PR which I merged make this one diverge.

Then I see another problem: you added one test case for your test, but you also added registry test file for case you don't own. IMHO, this is not your role to update these tests. Your role is more to handle the case where you don't reach that registry value.

So please remove also the following file from your PR and adapt the code so tests are not broken considering the registry key wasn't found:
1. `resources/win32/registry/amd-radeon-rx6600xt-6&EAF386A&0&00000019.reg`
2. `resources/win32/registry/intel+nvidia-3&11583659&0&10.reg`
3. `resources/win32/registry/intel+nvidia-4&1286464&0&0008.reg`
4. `resources/win32/registry/intel-hd-graphics-3&11583659&0&10.reg`
5. `resources/win32/registry/nvidia-geforce-rtx-2060-super-4&2F503515&0&0019.reg`

eduardomozart:
The `PCISLOT` test requires that these registry files exist. They were recreated following the `PCI` designation found on the existing test mocks, but they do not return `PCISLOT` because they are empty registry keys - they are kept only to avoid tests failing. The only way to avoid breaking the tests would be changing the `die` to `undef` on `t\lib\GLPI\Test\Utils.pm`, which would affect other tests.
I tried to rebase, but it seems to be removing the https://github.com/glpi-project/glpi-agent/commit/feb411f9c6e58e02d5e729f21346bf5d0678bede commit, which makes the tests fail. I'll take a look at it later.

eduardomozart (commented):
The tests are passing for me:
```
perl -Ilib -It/lib t/tasks/inventory/windows/controllers.t t/tasks/inventory/windows/video.t
1..3
ok 1 - microsoft-basic-display: parsing
ok 2 - microsoft-basic-display: registering
ok 3 - no warnings
```
It also passes the Q&A tests. Please also see the notes at https://github.com/glpi-project/glpi-agent/pull/1183#issuecomment-4671265261

g-bougard (changes requested):
I just pushed 1f8baae commit and it will have 2 sides effects on your PR:
1. You can now remove the 5 files this PR does no more need as mockGetRegistryKey() will return undef if file doesn't exist.
2. You have to rename `resources/win32/registry/microsoft-basic-display-3&11583659&0&10.reg` to `resources/win32/registry/microsoft-basic-display-3-11583659-0-10.reg`

For the second point, ampersand in a filename is not suitable as it has some meaning in some shell environment. So now we replace them by dash.

So can you rebase your PR against latest upstream develop branch and follow my 2 points ?

Also is it me or you removed a wmi dump for your "microsoft-basic-display" test case ? Was that intentional ?

eduardomozart (commented):
I updated the PR to the latest `develop` branch and removed the unused PCISLOT test mocks, also updated the PCISLOT mock filename. Tests are passing.
About the `microsoft-basic-display` WMI mock, it was merged previously into PR https://github.com/glpi-project/glpi-agent/pull/1181

g-bougard (approved):

PR #1184, opened 2026-06-08
feat(network): add network port statistics

eduardomozart:
### Description
This PR introduces network interface statistics collection, injecting traffic metrics directly into the existing `NETWORKS` node. It extracts hardware-level cumulative bytes and errors to allow accurate tracking, correlation, and monitoring of network performance.

image

> [!WARNING]
> **Important Note on GLPI Compatibility**
> This PR modifies the inventory format for `NETWORKS` nodes by appending the `IFINBYTES`, `IFOUTBYTES`, `IFINERRORS`, and `IFOUTERRORS` fields. 
> * **Server-Side Dependency:** This change is a companion to the server-side implementation currently proposed in **glpi-project/inventory_format/pull/174** and **glpi-project/glpi#24508**.
> * **Version Control:** To ensure compatibility with older GLPI servers, these fields are strictly gated behind a version check and will **only be transmitted if the target server is GLPI 12.0.0+**.

> [!NOTE]
> **Test Dependencies**
> For the tests in this PR to pass successfully, it requires `$interface->{DESCRIPTION}` to match the Windows connection name proposed in PR [glpi-project/glpi-agent#1176](https://github.com/glpi-project/glpi-agent/pull/1176) (already merged).

### Key Changes
* **Integrating stats into the `networks` node**: Added support to collect and report `IFINBYTES`, `IFOUTBYTES`, `IFINERRORS`, and `IFOUTERRORS` on network interface assets.
* **Cross-Platform Support**: 
  * **Windows**: Retrieves modern performance stats using `MSFT_NetAdapterStatisticsSettingData` via `StandardCimv2`, falling back to `Win32_PerfRawData_Tcpip_NetworkInterface` (PerSec counters as cumulative offsets) for older systems.
  * **Linux**: Parses cumulative interface stats dynamically from `/proc/net/dev`.
  * **macOS**: Parses cumulative hardware-layer link stats dynamically from `netstat -ib`.
* **Reliable Indexing**: Implements a sequential indexing (`_MODEL_COUNT`) mechanism for legacy Windows interfaces lacking GUIDs, securely mapping statistics to duplicate physical NICs.
* **GLPI 12+ Gating**: The statistics extraction logic across all operating systems requires `$inventory->supportsGlpiVersion('12.0.0')` to pass. 
* **Testing & Fixtures**: Added new unit tests across Windows, Linux, and macOS. The tests in `networks.t` have been unified and all tests are passing cleanly, including explicit validation proving that `addEntry()` seamlessly accepts the modified format.


eduardomozart:
Hello @g-bougard, on PR #24508, Trasher said he will wait for this to be committed before analyzing GLPI side code changes.

g-bougard:
I'll take an eye on it after next release.

Anyway, any format update will also require a dedicated [inventory_format spec project](https://github.com/glpi-project/inventory_format) PR.

eduardomozart:
This PR does not require modification to the inventory_schema; it uses the existing `network_ports` entry used by NetworkEquipment device classes ported to Computer assets.

lib/GLPI/Agent/Task/Inventory/Win32/Networks.pm
```diff
@@ -22,6 +22,42 @@ sub doInventory {
         or return;
 
     my $inventory = $params{inventory};
+
+    my %statistics;
+
+    if ($inventory->{_glpi_version} >= glpiVersion('12')) {
```
g-bougard: Inventory object now has a method to check glpiVersion (recently added by another contributor PR):
```suggestion
    if ($inventory->supportsGlpiVersion('12')) {
```

lib/GLPI/Agent/Task/Inventory/Win32/Networks.pm
```diff
@@ -22,6 +22,42 @@ sub doInventory {
...
+        my @modern_stats = getWMIObjects(
+            moniker    => 'winmgmts://./root/StandardCimv2',
+            class      => 'MSFT_NetAdapterStatisticsSettingData',
+            properties => [ qw/Name ReceivedBytes SentBytes ReceivedPacketErrors OutboundPacketErrors/ ]
+        );
+
+        foreach my $stat (@modern_stats) {
+            $statistics{$stat->{Name}} = {
+                ifinoctets  => $stat->{ReceivedBytes},
+                ifoutoctets => $stat->{SentBytes},
```
g-bougard: Use `bytes` word:
```suggestion
                ifinbytes  => $stat->{ReceivedBytes},
                ifoutbytes => $stat->{SentBytes},
```

lib/GLPI/Agent/Task/Inventory/Win32/Networks.pm
```diff
@@ -22,6 +22,42 @@ sub doInventory {
...
+            class      => 'Win32_PerfRawData_Tcpip_NetworkInterface',
+            properties => [ qw/Name BytesReceivedPersec BytesSentPersec PacketsReceivedErrors PacketsOutboundErrors/ ]
+        );
+
+        foreach my $stat (@legacy_stats) {
+            # Keep modern stats if they exist (they map nicely by DESCRIPTION)
+            # But populate legacy ones too for adapters that only show up here
+            $statistics{$stat->{Name}} //= {
+                ifinoctets  => $stat->{BytesReceivedPersec},
+                ifoutoctets => $stat->{BytesSentPersec},
```
g-bougard: Use `bytes` word:
```suggestion
                ifinbytes  => $stat->{BytesReceivedPersec},
                ifoutbytes => $stat->{BytesSentPersec},
```
But using stat "Persec" here is probably an error.

lib/GLPI/Agent/Task/Inventory/Win32/Networks.pm
```diff
@@ -31,7 +67,20 @@ sub doInventory {
...
+    my %model_counts;
     foreach my $interface (@interfaces) {
+        my $lookup_name = $interface->{MODEL} || '';
+        if ($lookup_name) {
+            $interface->{_MODEL_COUNT} = ++$model_counts{$lookup_name};
+            if ($interface->{_MODEL_COUNT} > 1) {
+                $lookup_name .= ' _' . $interface->{_MODEL_COUNT};
+            }
+        }
+
```
g-bougard: `$lookup_name` is only used in glpi 12 is condition below. This code should be moved there.

lib/GLPI/Agent/Task/Inventory/Win32/Networks.pm
```diff
@@ -47,6 +96,31 @@ sub doInventory {
...
+                # We track them by MAC and DESCRIPTION so we only inject one NETWORKPORTS block per physical card.
+                my $seen_key = "seen_" . ($interface->{MACADDR} || '') . "_" . ($interface->{DESCRIPTION} || '');
+                if (!$statistics{$seen_key}) {
+                    $statistics{$seen_key} = 1;
+                    my $network_port = {
+                        NAME        => $interface->{DESCRIPTION},
+                        MAC         => $interface->{MACADDR},
+                        IFNUMBER    => $interface->{_IFNUMBER},
+                        IFINOCTETS  => $stat->{ifinoctets},
+                        IFOUTOCTETS => $stat->{ifoutoctets},
```
g-bougard: Use "bytes" word here, "octets" is not part of the inventory_format specs as far I know.
```suggestion
                        IFINBYTES  => $stat->{ifinbytes},
                        IFOUTBYTES => $stat->{ifoutbytes},
```

lib/GLPI/Agent/Inventory.pm
```diff
@@ -65,6 +65,7 @@ my %fields = (
     MODEMS           => [ qw/DESCRIPTION NAME TYPE MODEL/ ],
     MONITORS         => [ qw/BASE64 CAPTION DESCRIPTION MANUFACTURER SERIAL
                              UUENCODE NAME TYPE ALTSERIAL PORT/ ],
+    NETWORK_PORTS    => [ qw/NAME MAC IFNUMBER IFINOCTETS IFOUTOCTETS IFINERRORS IFOUTERRORS/ ],
```
g-bougard: Use `bytes` word:
```suggestion
    NETWORK_PORTS    => [ qw/NAME MAC IFNUMBER IFINBYTES IFOUTBYTES IFINERRORS IFOUTERRORS/ ],
```

t/tasks/inventory/windows/network_ports.t
```diff
@@ -0,0 +1,117 @@
...
+package main;
+
+my %tests = (
+    'loadbalance' => [
+        {
+            MAC => '20:47:47:90:78:42',
+            NAME => 'Ethernet',
+            IFNUMBER => 13,
+            IFINOCTETS => 1412918786473,
+            IFOUTOCTETS => 364462371847,
```
g-bougard: Use `bytes` word

t/tasks/inventory/windows/network_ports.t
```diff
@@ -0,0 +1,117 @@
...
+            IFINERRORS => 0,
+            IFOUTERRORS => 0
+        },
+        {
+            MAC => '20:47:47:90:78:44',
+            NAME => 'Ethernet 2',
+            IFNUMBER => 8,
+            IFINOCTETS => 70987676,
+            IFOUTOCTETS => 80987676,
```
g-bougard: Use `bytes` word

t/tasks/inventory/windows/network_ports.t
```diff
@@ -0,0 +1,117 @@
...
+            IFINERRORS => 0,
+            IFOUTERRORS => 0
+        },
+        {
+            MAC => '20:47:47:90:78:42',
+            NAME => 'LoadBalance',
+            IFNUMBER => 4,
+            IFINOCTETS => 1403973923755,
+            IFOUTOCTETS => 726370059830,
```
g-bougard: Use `bytes` word

g-bougard (changes requested):
Hi @eduardomozart 

this is a first review where I see few problems.

To be honest, I don't like to add "network_ports" node support on computers:
 - few data like "name" and "mac" are duplicated
 - "networks" node still reports network interfaces
 - "network_ports" node format was implemented for networking device with a really different scope.

For me, it would be more accurate to update format to include desired statistics directly into "network" nodes. And it would probably more easy to integrate into GLPI.

In the windows stats you use for the older case, you get "PerSec" stats, I don't think this is the correct one as it means you use a different unit than networking devices and even than the modern statistics.

Can you better study another way of doing by updating "networks" inventory format to include statistics and so insert stats directly in "networks" node ?

eduardomozart (commented):
Hi @g-bougard, 

Thanks for the review and the feedback! I've refactored the approach based on your suggestions:

1. **Integrating stats into the `networks` node**: I've dropped the `network_ports` approach entirely. I've updated the script to inject the statistics (`ifinbytes`, `ifoutbytes`, `ifinerrors`, `ifouterrors`) directly into the `NETWORKS` node, taking care to ensure these new statistics are sent only if the server is GLPI 12.0.0 or higher. The tests in `networks.t` have also been unified and all tests are passing cleanly. To support this natively, I have opened a PR updating the inventory format schema over at: https://github.com/glpi-project/inventory_format/pull/174. Additionally, https://github.com/glpi-project/glpi/pull/24508 has also been updated to reflect the changes on the server side.

2. **The `PerSec` counter units**: You raise a great point about the naming. While they do end in `PerSec` (`BytesReceivedPersec`, `BytesSentPersec`), they are being pulled from the `Win32_PerfRawData_*` class. In Windows WMI, the *RawData* performance classes actually expose the raw, cumulative totals since boot. It is only when querying the *FormattedData* variant of that class that WMI calculates the per-second delta. Despite Microsoft's confusing property naming, these values are indeed the absolute cumulative bytes—matching the modern `MSFT_NetAdapter` statistics perfectly. I have added a comment directly in the code to clarify this quirk for future maintainers.

lib/GLPI/Agent/Task/Inventory/Win32/Networks.pm
```diff
@@ -31,7 +70,21 @@ sub doInventory {
...
     foreach my $interface (@interfaces) {
+        my $lookup_base = $interface->{MODEL} || $interface->{DESCRIPTION} || '';
+        my $lookup_name = $lookup_base;
+        if ($lookup_base) {
+            $interface->{_MODEL_COUNT} = ++$model_counts{$lookup_base};
+            if ($interface->{_MODEL_COUNT} > 1) {
+                $lookup_name .= ' _' . $interface->{_MODEL_COUNT};
+            }
+        }
+
```
g-bougard: As I said in my previous review, you should move you added lines after line 103 under the GLPI v12 support test and just before `$lookup_name` is really used.

t/tasks/inventory/windows/networks.t
```diff
@@ -25,6 +28,17 @@ Test::NoWarnings->use();
...
+sub new { bless { sections => {}, _glpi_version => 12_000_000 }, shift }
+sub supportsGlpiVersion { 1 }
+sub addEntry {
+    my ($self, %params) = @_;
+    push @{$self->{sections}->{$params{section}}}, $params{entry};
+}
+sub setHardware { }
+
+package main;
+
```
g-bougard: Sorry, but this code is horrible.

Better setup a normal inventory object with the parameters you need.

About setHardware and addEntry, you should better split the code under doInventory() to have a _getInterfaces() in lib/GLPI/Agent/Task/Inventory/Win32/Networks.pm as this is the case for many other cases. And this will help to simplify unittest code.

t/tasks/inventory/windows/networks.t
```diff
@@ -36,10 +50,43 @@ my %tests = (
...
+            MODEL => 'Microsoft Network Adapter Multiplexor Driver',
+            DESCRIPTION => 'LoadBalance',
+            IFINBYTES => 1403973923755,
+            IFOUTBYTES => 726370059830,
+            IFINERRORS => 0,
+            IFOUTERRORS => 0
+        })
+    ]
+);
+
```
g-bougard, line 100: I don't agree to limit the tests to expected statistics on expected port using superhashof() and also using superbagof below. Please, better check a _getInterfaces() method result with a simple cmp_deeply() call.

And indeed this was a missing here.

g-bougard (changes requested):
Hi @eduardomozart 

thank you for that update.

I now mostly agree with your code. But you missed to follow one of my comment about $lookup_name compute, see below.

And mostly, I don't agree with the test code. Check my comment and do what you can. If you don't understand what I mean, at least fix the $lookup_name compute code location and I'll handle the test part merging manually.

g-bougard:
Running tests locally, I also find an issue with t/agent/tools/win32.t test. It's related to the change you made on lib/GLPI/Agent/Tools/Win32/NetAdapter.pm to include a temporary _IFNUMBER value.

eduardomozart (commented):
Hi @g-bougard,

I have pushed a new commit to address your concerns:

1. **`$lookup_name` logic scope**: I've moved the `_MODEL_COUNT` and `$lookup_name` generation logic down into the `if ($inventory->supportsGlpiVersion('12.0.0'))` block, right before it is actually used.
2. **`Networks.pm` refactoring**: I've split `doInventory()` and extracted the interface processing logic into a `_getInterfaces($inventory)` method. `doInventory()` now just iterates over its result to populate `addEntry` and `setHardware`.
3. **`networks.t` test improvements**: I've removed the `MockInventory` package hack completely. The test now natively initializes `GLPI::Agent::Inventory->new(glpi => '12.0.0')` and tests the direct output of `_getInterfaces()`.
4. **Strict test assertions**: I've removed the `superhashof()` and `superbagof()` usage in the tests. It now relies on a strict `cmp_deeply()` array comparison against the exact expected properties for each network interface.
5. **Fixed `win32.t`**: I realized `_IFNUMBER` wasn't even strictly necessary for our logic, so I've completely removed it from `NetAdapter.pm`. All tests across the suite are now passing cleanly!

lib/GLPI/Agent/Task/Inventory/Win32/Networks.pm
```diff
@@ -113,20 +125,14 @@ sub doInventory {
...
             }
+            delete $interface->{_MODEL_COUNT};
         }
-        delete $interface->{_MODEL_COUNT};
-        delete $interface->{_IFNUMBER};
-        $inventory->addEntry(
-            section => 'NETWORKS',
-            entry   => $interface
-        );
+        delete $interface->{dns};
```
g-bougard: This is not the right place to cleanup dns field.

It has to be done in doInventory(). For me you're breaking HARDWARE/DNS inventory.

t/tasks/inventory/windows/networks.t
```diff
@@ -28,16 +28,7 @@ Test::NoWarnings->use();
...
-sub new { bless { sections => {}, _glpi_version => 12_000_000 }, shift }
-sub supportsGlpiVersion { 1 }
-sub addEntry {
-    my ($self, %params) = @_;
-    push @{$self->{sections}->{$params{section}}}, $params{entry};
-}
-sub setHardware { }
-
-package main;
+use GLPI::Agent::Inventory;
```
g-bougard: Move it next to `use GLPI::Test::Utils;` line 14.

t/tasks/inventory/windows/networks.t
g-bougard: Just `use Test::Deep;` here.

lib/GLPI/Agent/Task/Inventory/Win32/Networks.pm
```diff
@@ -18,14 +18,38 @@ sub isEnabled {
...
+
+    $inventory->setHardware({
+        DEFAULTGATEWAY => join('/', uniq @gateways),
+        DNS            => join('/', uniq @dns),
+    });
+
+}
+
+sub _getInterfaces {
+    my ($inventory) = @_;
```
g-bougard: Use %params syntax here, this is more a convention. But you don't need to pass $inventory if you precompute a glpi12_support field.

lib/GLPI/Agent/Task/Inventory/Win32/Networks.pm
```diff
@@ -18,14 +18,38 @@ sub isEnabled {
 sub doInventory {
     my (%params) = @_;
 
+    my $inventory = $params{inventory};
+
+    my (@gateways, @dns);
+
+    foreach my $interface (_getInterfaces($inventory)) {
```
g-bougard: Better pass a hash to _getInterfaces() and you can set glpi12_support field at the same time:
```
    foreach my $interface (_getInterfaces(
        glpi12_support => $inventory->supportsGlpiVersion('12.0.0')
    )) {
```

lib/GLPI/Agent/Task/Inventory/Win32/Networks.pm
```diff
@@ -18,14 +18,38 @@ sub isEnabled {
...
+
     my @interfaces = getInterfaces()
         or return;
 
-    my $inventory = $params{inventory};
-
     my %statistics;
 
-    if ($inventory->supportsGlpiVersion('12')) {
+    if ($inventory && $inventory->supportsGlpiVersion('12.0.0')) {
```
g-bougard: If you follow my previous comment, use a condition like:
```
    if ($params{glpi12_support}) {
```
Also update test to pass the expected value.

lib/GLPI/Agent/Task/Inventory/Win32/Networks.pm
```diff
@@ -61,46 +85,34 @@ sub doInventory {
...
-        delete $interface->{DNSDomain};
-        delete $interface->{GUID};
-
         if ($interface->{PNPDEVICEID} && !$interface->{TYPE}) {
             my $type = _getMediaType($interface->{PNPDEVICEID}, $keys);
             $interface->{TYPE} = $type if defined($type);
         }
 
-        if ($inventory->supportsGlpiVersion('12.0.0')) {
+        if ($inventory && $inventory->supportsGlpiVersion('12.0.0')) {
```
g-bougard: If you follow my previous comment, use a condition like:
```
        if ($params{glpi12_support}) {
```

t/tasks/inventory/windows/networks.t
```diff
@@ -127,19 +143,15 @@ foreach my $test (keys %network_ports_tests) {
...
-
-    my $ports = $inventory->{sections}->{NETWORKS} || [];
-
     cmp_deeply(
-        $ports,
-        superbagof(@{$network_ports_tests{$test}}),
+        \@ports,
+        $network_ports_tests{$test},
         "$test sample NETWORKS matches expected counters"
     );
```
g-bougard, line 162: As done in many places in glpi-agent unittests, and as this is what we expect to be done un doInventory make a test where we don't generate an exception when using this code for each port:
```
        $inventory->addEntry(
            section => 'NETWORKS',
            entry   => $interface
        );
```
For this specific test, you'll may have to clean up "dns" field from each port.

This test is intended to prove we don't break the expected inventory format.

lib/GLPI/Agent/Task/Inventory/Win32/Networks.pm
```diff
@@ -61,46 +85,34 @@ sub doInventory {
...
+            # The legacy WMI class Win32_PerfRawData_Tcpip_NetworkInterface lacks strict linkage properties (like MAC or GUID).
+            # When multiple physical NICs of the exact same model exist, Windows natively assigns them sequential suffixes
+            # (e.g. "_2", " _3") based on their PnP enumeration order. Since getInterfaces natively arrays them
+            # in this exact same index order, we generate sequential lookup strings here to map them 1:1 reliably.
+            my $lookup_base = $interface->{MODEL} || $interface->{DESCRIPTION} || '';
+            my $lookup_name = $lookup_base;
+            if ($lookup_base) {
+                $interface->{_MODEL_COUNT} = ++$model_counts{$lookup_base};
```
g-bougard: Do we really need to use a private interface data here ? It seems only used to update lookup_name and is delete just after.

Can you validate if just using such code is correct ?
```
            my $model_count = ++$model_counts{$lookup_base};
            $lookup_name .= ' _' . $model_count if $model_count > 1;
```
And if it is sufficient, just remove also the `delete $interface->{_MODEL_COUNT};` line below.

g-bougard (changes requested):
Thank you, it's better.

But I still few problems.

Can you review my comments ?

eduardomozart (commented):
Hi @g-bougard,

I've updated the PR with all of your latest suggestions:

1. **DNS cleanup placement**: I reverted the cleanup of the `dns`, `DNSDomain`, and `GUID` fields from `_getInterfaces()` back into `doInventory()`. That ensures we don't accidentally break the HARDWARE/DNS inventory mapping.
2. **`%params` pattern**: Changed `_getInterfaces` to accept `%params` and we now pass `glpi12_support => $inventory->supportsGlpiVersion('12.0.0')` when calling it from `doInventory()`.
3. **`_MODEL_COUNT` cleanup**: I used your exact suggestion here. The inline `$model_count` logic looks much cleaner and entirely avoids the need to inject (and subsequently delete) the private data into the interface hash.
4. **`networks.t` updates**:
    - Moved the `GLPI::Agent::Inventory` use statement to line 14.
    - Updated `Test::Deep` import to avoid `superbagof`/`superhashof` bloat.
    - Added the explicit loop running `$inventory->addEntry(...)` for each returned port to explicitly prove that our new structure is valid and doesn't throw exceptions.

All unit tests are passing cleanly.

lib/GLPI/Agent/Task/Inventory/Linux/Networks.pm
```diff
@@ -131,12 +153,19 @@ sub _getInterfaces {
             );
         }
 
+        if (!$interface->{MTU} && canRead("/sys/class/net/$interface->{DESCRIPTION}/mtu")) {
+            my $mtu = getFirstLine(
+                file => "/sys/class/net/$interface->{DESCRIPTION}/mtu"
```
g-bougard: logger params should be used for debugging
```suggestion
                file   => "/sys/class/net/$interface->{DESCRIPTION}/mtu",
                logger => $logger
```

lib/GLPI/Agent/Task/Inventory/Linux/Networks.pm
```diff
@@ -131,12 +153,19 @@ sub _getInterfaces {
...
+            );
+            $interface->{MTU} = $mtu if $mtu && $mtu =~ /^\d+$/ && $mtu > 0;
```
g-bougard: last test is redundant:
 * first `$mtu` condition validates `$mtu` is defined for the following regex check, but it also validates it evaluates to a number different than zero
 * second `$mtu =~ /^\d+$/` also validates `$mtu` is positive
So `$mtu > 0` here will always be true as third check:
```suggestion
            $interface->{MTU} = $mtu if $mtu && $mtu =~ /^\d+$/;
```

lib/GLPI/Agent/Task/Inventory/Linux/Networks.pm
```diff
@@ -131,12 +153,19 @@ sub _getInterfaces {
...
+        }
+
         if (defined($interface->{STATUS}) && $interface->{STATUS} eq 'Up') {
             if (canRead("/sys/class/net/$interface->{DESCRIPTION}/speed")) {
                 my $speed = getFirstLine(
                     file => "/sys/class/net/$interface->{DESCRIPTION}/speed"
                 );
-                $interface->{SPEED} = $speed && $speed > 0 ? $speed : 0;
+                $interface->{SPEED} = ($speed && $speed =~ /^\d+$/ && $speed > 0 && $speed != SPEED_UNKNOWN && $speed != SPEED_UNKNOWN_32) ? $speed : 0;
```
g-bougard: Here also third test is redundant:
```suggestion
                $interface->{SPEED} = ($speed && $speed =~ /^\d+$/ && $speed != SPEED_UNKNOWN && $speed != SPEED_UNKNOWN_32) ? $speed : 0;
```

lib/GLPI/Agent/Task/Inventory/Linux/Networks.pm
```diff
@@ -54,6 +57,25 @@ sub _getInterfaces {
...
+                my @values = split(/\s+/, trimWhitespace($2));
+                # Format:
+                # RX: bytes(0), packets(1), errs(2), drop(3), fifo(4), frame(5), compressed(6), multicast(7)
+                # TX: bytes(8), packets(9), errs(10), drop(11), fifo(12), colls(13), carrier(14), compressed(15)
+                $statistics{$name} = {
+                    ifinbytes  => $values[0],
+                    ifinerrors  => $values[2],
+                    ifoutbytes => $values[8],
+                    ifouterrors => $values[10],
+                };
```
g-bougard: You should simplify using a regex to still extract only useful stats like you also does in MacOS version:
```suggestion
            if ($line =~ /^\s*([^:]+):\s*(\d+)\s+\d+\s+(\d+)\s+\d+\s+\d+\s+\d+\s+\d+\s+\d+\s+(\d+)\s+\d+\s+(\d+)/) {
                $statistics{$1} = {
                    ifinbytes  => $2,
                    ifinerrors  => $3,
                    ifoutbytes => $4,
                    ifouterrors => $5,
                };
```

lib/GLPI/Agent/Task/Inventory/MacOS/Networks.pm
```diff
@@ -54,7 +57,58 @@ sub _getInterfaces {
...
+        foreach my $line (@netstat_lines) {
+            # Looking for Link layer line: Name Mtu Network Address Ipkts Ierrs Ibytes Opkts Oerrs Obytes Coll
+            if ($line =~ /^(\S+)\s+\d+\s+<Link#\d+>\s+(?:(?:[a-fA-F0-9:]+:[a-fA-F0-9:]+)\s+)?(\d+)\s+(\d+)\s+(\d+)\s+(\d+)\s+(\d+)\s+(\d+)/) {
+                $statistics{$1} = {
+                    ifinerrors  => $3,
+                    ifinbytes   => $4,
+                    ifouterrors => $6,
+                    ifoutbytes  => $7,
+                };
+            }
```
g-bougard: You can avoid few group matching and simplify your mac address matching:
```suggestion
            if ($line =~ /^(\S+)\s+\d+\s+<Link#\d+>\s+[a-fA-F0-9:]+\s+\d+\s+(\d+)\s+(\d+)\s+\d+\s+(\d+)\s+(\d+)/) {
                $statistics{$1} = {
                    ifinerrors  => $2,
                    ifinbytes   => $3,
                    ifouterrors => $4,
                    ifoutbytes  => $5,
                };
            }
```

lib/GLPI/Agent/Task/Inventory/MacOS/Networks.pm
```diff
@@ -159,8 +214,8 @@ sub _parseIfconfig {
         if ($line =~ /media (\S+)/ && empty($interface->{TYPE})) {
             $interface->{TYPE} = $1;
         }
-        if ($line =~ /media: \S+ \((\d+)baseTX <.*>\)/) {
-            $interface->{SPEED} = $1;
+        if ($line =~ /media: \S+ \((?:(\d+)G)?(\d+)?base[^ ]* <.*>\)/i) {
+            $interface->{SPEED} = $1 ? $1 * 1000 : $2;
```
g-bougard: I don't understand this change in regex et SPEED computing. Have you a test case ?
It may also been interesting to re-use getCanonicalInterfaceSpeed() api if this is possible.

lib/GLPI/Agent/Task/Inventory/Win32/Networks.pm
```diff
@@ -18,20 +18,13 @@ sub isEnabled {
...
     my $inventory = $params{inventory};
-    my (@gateways, @dns);
 
-    my $keys;
-    $keys = getRegistryKey(
-        path   => "HKEY_LOCAL_MACHINE/SYSTEM/CurrentControlSet/Control/Network/{4D36E972-E325-11CE-BFC1-08002BE10318}",
-        # Important for remote inventory optimization
-        required    => [ qw/PnpInstanceID MediaSubType/ ],
-    ) if grep { $_->{PNPDEVICEID}} @interfaces;
+    my (@gateways, @dns);
```
g-bougard, line 23: You should remove empty line just before as there's here indeed no change regarding lines 24 && 25.

lib/GLPI/Agent/Tools/Linux.pm
```diff
@@ -544,6 +547,9 @@ sub getInterfacesFromIfconfig {
             }
 
         }
+        if ($line =~ /MTU:?(\d+)/i || $line =~ /mtu (\d+)/i) {
```
g-bougard: To optimize, you can use here only one regex:
```suggestion
        if ($line =~ /(?:MTU:?|mtu )(\d+)/i) {
```

t/tasks/inventory/linux/networks.t
```diff
@@ -23,10 +25,54 @@ my %tests = (
...
+my $inventory = GLPI::Agent::Inventory->new(glpi => '12.0.0');
+eval {
+    foreach my $port (@interfaces) {
+        $inventory->addEntry(
+            section => 'NETWORKS',
+            entry   => $port
+        );
+    }
+};
+ok(!$@, "addEntry() doesn't throw exceptions for Linux networks");
```
g-bougard: Testing an eval result is not the way we use in general in our tests, try this:
```suggestion
lives_ok {
    foreach my $port (@interfaces) {
        $inventory->addEntry(
            section => 'NETWORKS',
            entry   => $port
        );
    }
} "addEntry() doesn't throw exceptions for Linux networks";
```

t/tasks/inventory/macos/networks.t
```diff
@@ -782,12 +790,30 @@ foreach my $test (keys %tests) {
...
+        my $inventory = GLPI::Agent::Inventory->new(glpi => '12.0.0');
+        eval {
+            foreach my $port (@{$nets}) {
+                $inventory->addEntry(
+                    section => 'NETWORKS',
+                    entry   => $port
+                );
+            }
+        };
+        ok(!$@, "addEntry() doesn't throw exceptions for $test");
```
g-bougard: Better try this:
```suggestion
        lives_ok {
            foreach my $port (@{$nets}) {
                $inventory->addEntry(
                    section => 'NETWORKS',
                    entry   => $port
                );
            }
        } "addEntry() doesn't throw exceptions for $test";
```

t/tasks/inventory/windows/networks.t
```diff
@@ -55,3 +117,58 @@ foreach my $test (keys %tests) {
...
+    my $inventory = GLPI::Agent::Inventory->new(glpi => '12.0.0');
+    eval {
+        foreach my $port (@ports) {
+            $inventory->addEntry(
+                section => 'NETWORKS',
+                entry   => $port
+            );
+        }
+    };
+    is($EVAL_ERROR, '', "addEntry does not throw exceptions for $test ports");
```
g-bougard: Better try this:
```suggestion
    lives_ok {
        foreach my $port (@ports) {
            $inventory->addEntry(
                section => 'NETWORKS',
                entry   => $port
            );
        }
    } "addEntry does not throw exceptions for $test ports";
```

g-bougard (changes requested):
Hi @eduardomozart 

thank you to also include MacOS and linux cases.

Now we are next from a merging. I only see few trivial requried changes.

There's also one change I don't understand on SPEED computing on MacOSX. Can you check my comment on that case and clarify that point ?

lib/GLPI/Agent/Task/Inventory/MacOS/Networks.pm
eduardomozart: Hi @g-bougard,

The original reason for that regex is that on macOS, `ifconfig` doesn't report network speed on an isolated property line. Instead, it embeds the speed right inside the `media` string. For example, Ethernet connections often show up as:
`media: autoselect (1000baseT <full-duplex,flow-control>)`

Since the speed information was technically present in the raw output, we were previously missing out on extracting it!

To make `getCanonicalInterfaceSpeed()` work seamlessly, I've pushed a new commit that updates `getCanonicalInterfaceSpeed` in `GLPI::Agent::Tools` so that it now natively understands the `ba` (base) suffix. 

Now, we simply parse out that inner string (e.g. `1000baseT` or `10Gbase-T`) and pass it directly into `getCanonicalInterfaceSpeed($1)` which cleanly handles the rest!

As for test cases, this is fully covered in `t/tasks/inventory/macos/networks.t` using the existing mock `ifconfig` output files in `resources/macos/ifconfig/`. For example, the `macbook-pro-16-intel-2019-catalina` mock explicitly uses `1000baseT` and tests for `SPEED => 1000`. All unit tests are passing cleanly with this new refactored logic!

eduardomozart (commented):
Done.

g-bougard:
There's an error in `t/tasks/inventory/linux/networks.t`.

I'm going to merge the PR and I'll fix the test myself just after as it needs some more refacto.

eduardomozart:
Sorry, I tried to commit the fix, but it was merged in the middle of the process.

```diff
commit bdbf641019b223a54cdace09cb171714456ab9d2
Author: E <2974895+eduardomozart@users.noreply.github.com>
Date:   Fri Jul 31 13:09:08 2026 -0300
    Fix Test::MockModule original method usage and missing Test::Exception in Linux networks test
diff --git a/t/tasks/inventory/linux/networks.t b/t/tasks/inventory/linux/networks.t
index 646f98146..631d62d81 100644
--- a/t/tasks/inventory/linux/networks.t
+++ b/t/tasks/inventory/linux/networks.t
@@ -7,6 +7,7 @@ use lib 't/lib';
 use Test::Deep;
 use Test::More;
 use Test::NoWarnings;
+use Test::Exception;
 
 use GLPI::Agent::Task::Inventory::Linux::Networks;
 use GLPI::Agent::Inventory;
@@ -36,7 +37,6 @@ foreach my $test (keys %tests) {
 my $linux_net = Test::MockModule->new('GLPI::Agent::Task::Inventory::Linux::Networks');
 $linux_net->mock('_getInterfacesBase', sub { return ({ DESCRIPTION => 'eth0' }); });
 
-my $orig_getAllLines = $linux_net->original('getAllLines');
 $linux_net->mock('getAllLines', sub {
     my (%params) = @_;
     if ($params{file} && $params{file} eq '/proc/net/dev') {
@@ -46,7 +46,7 @@ $linux_net->mock('getAllLines', sub {
             '  eth0: 1234567   12345    1    0    0     0          0         0 8765432   87654    2    0    0     0       0          0'
         );
     }
-    return $orig_getAllLines->(@_);
+    return $linux_net->original('getAllLines')->(@_);
 });
 
 my @interfaces = GLPI::Agent::Task::Inventory::Linux::Networks::_getInterfaces(glpi12_support => 1);
 ```

g-bougard:
Yes, but I wanted to make more refacto to be able to add more tests. Here is my final udpate: afe1c80

I used your data and merged it we some other to obtain a more accurate test and being able to add more. This is also a base to permit testing more advanced cases.

PR #1189, opened 2026-06-11
added support for eltex MES2348B 52-port 1G/10G

hlebuschek:


hlebuschek:
I'd also like to point out a performance concern related to the agent's architecture. As far as I can tell, during network discovery/inventory the agent reads the SNMP walk one line (one OID) at a time rather than processing it in bulk. On a device that returns this many OIDs, that line-by-line approach makes the scan very slow.

  It would be great if the walk could be parsed/consumed in larger batches (or streamed in chunks) to significantly reduce the processing time on devices with a large number of OIDs like this Eltex switch.
[walk_eltex_full_bulk.txt.zip](https://github.com/user-attachments/files/28909795/walk_eltex_full_bulk.txt.zip)

g-bougard (changes requested):
Hello @hlebuschek 

I see only one thing which is not correct:
1. `share/sysobject.ids` file change should be done on [glpi-project/sysobject.ids project](https://github.com/glpi-project/sysobject.ids)

After it will be merged, it will be automatically updated in glpi-agent during release process.

So you have for this sysobject.ids file change to:
1. Remove it from this PR
2. Open a PR on [glpi-project/sysobject.ids project](https://github.com/glpi-project/sysobject.ids)

Also as v1.18 version was released few weeks ago, the Changes diff is no more correct. So can you rebase against current develop branch ?

g-bougard:
Hi @hlebuschek 

thank you for your contribution. It will be included in next nightly build.

I forgot to comment your suggestion about the "performance concern related to the agent's architecture". As far I know, we choose to keep agent conservative as some old devices just doesn't support bulk requests. Anyway, I think we may try to enhance something around this. Maybe we can test the device comparing the result of a bulk request to the result on each request. Maybe we can assume bulk requests are supported in some case like if snmp v3 is used.

Anyway, the best by now is to open a dedicated issue to discuss about it with the community. Can I leave you open this issue ?

PR #1217, opened 2026-08-02
Feature/apk software inventory

Megachip:
# Add software inventory support for Alpine Package Keeper (apk)

Closes the request in discussion #901.

## Why

Alpine Linux is the default distribution for a large share of Docker base
images, but glpi-agent had no software-inventory support for its package
manager `apk`. Installed apk packages were therefore missing from the
inventory entirely.

## What

New module `GLPI::Agent::Task::Inventory::Generic::Softwares::APK`.

It reads the apk **installed packages database** (`/lib/apk/db/installed`)
directly rather than forking the `apk` binary. This was chosen over parsing
`apk list --installed` output because it is:

- **faster** — a single file read instead of spawning `apk`, which parses that
  very same file itself;
- **more robust** — the package name and version are stored in **separate
  fields** (`P:` / `V:`), so there is no fragile `name-version` token splitting
  (apk names may contain digits, e.g. `libcrypto3`, which makes heuristic
  splitting error-prone). This also sidesteps the apk v2/v3 `list` output
  differences raised in the discussion;
- **richer** — installed size and description are available in the same pass.

The database describes each package as a block of `key:value` lines separated
by a blank line. The module extracts:

| apk field | GLPI field |
|-----------|------------|
| `P` | `NAME` |
| `V` | `VERSION` |
| `A` | `ARCH` |
| `I` | `FILESIZE` (installed size, in bytes) |
| `T` | `COMMENTS` (description) |
| `m` | `PUBLISHER` (maintainer, when present) |
| — | `FROM => 'apk'` |

`isEnabled()` requires both `canRun('apk')` and the database file to be present.
Maintainer names are UTF-8 decoded (like the RPM module). `INSTALLDATE` is
deliberately not set: the database only stores the package build time (`t:`),
not a per-host installation date (the Deb module omits it for the same reason).

## Tests

- Added fixture `resources/linux/packaging/apk` (a realistic `installed`
  database, incl. a multi-hyphen name, a digit-suffixed name `libcrypto3`, a
  `noarch` package and one package without a maintainer field).
- Extended `t/tasks/inventory/linux/softwares.t` with apk parsing and
  registration checks. Full file passes (27/27). `use utf8` was added so the
  accented maintainer names in the expectations compare correctly.

## Notes

- `opkg` is intentionally out of scope here and can follow in a separate PR.
- apk-tools 3 will move the installed database to a binary ADB format; when that
  reaches stable Alpine, the parser will need a format check. All currently
  deployed Alpine releases use the text format handled here.


g-bougard (approved):

g-bougard:
Thank you for your contribution.

PR #1218, opened 2026-08-03
Improve sysread handling in Daemon.pm

phillscholl:
Handle EOF and errors in sysread for child process. Issue #1041

### DIFF
--- a/lib/GLPI/Agent/Daemon.pm
+++ b/lib/GLPI/Agent/Daemon.pm
@@ -532,7 +532,14 @@ sub handleChildren {
         delete $child->{in} unless $child->{in} && $child->{in}->opened;
         while ($child->{in} && $child->{pollin} && $child->{poll} && &{$child->{poll}}($child->{pollin})) {
             my $msg = " " x 5;
-            if ($child->{in}->sysread($msg, 5)) {
+            my $read = $child->{in}->sysread($msg, 5);
+            unless ($read) {
+                # 0 == EOF (peer closed), undef == error. The descriptor stays  
+                # readable forever, so stop polling it and forget the child. 
+                $self->child_exit($pid);
+                last;
+           }
+           if ($read) {
                 if ($msg eq IPC_LEAVE) {
                     $self->child_exit($pid);
                 } elsif ($msg eq IPC_RNAME) {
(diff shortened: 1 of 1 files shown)


lib/GLPI/Agent/Daemon.pm
```diff
@@ -532,7 +532,14 @@ sub handleChildren {
...
             my $msg = " " x 5;
-            if ($child->{in}->sysread($msg, 5)) {
+            my $read = $child->{in}->sysread($msg, 5);
+            unless ($read) {
+                # 0 == EOF (peer closed), undef == error. The descriptor stays  
+                # readable forever, so stop polling it and forget the child. 
+                $self->child_exit($pid);
+                last;
+           }
+           if ($read) {
```
g-bougard: Can you better keep the original `if` and add a "else" block to have the following starting from l. 576:
```perl
           } else {
                # 0 == EOF (peer closed), undef == error. The descriptor stays
                # readable forever, so stop polling it and forget the child.
                $self->child_exit($pid);
           }
           $count++;
```
The `last` command is not required as we still can continue to communicate with other ready childs... and eventually stop polling other failing ones.

g-bougard (changes requested):
Hi @phillscholl 

to be honest, and as you know, I'm suspicious about this change. I don't know if this can also close a child which triggered polling but didn't send anything and still would later.

And I'm suspicious as it seems it happens only on some Apple hardware. I'm anyway not sure the meaning you describe is correct. And we don't have a hardware on which we were able to reproduce. 
Maybe the child_exit call should be done only on MacOSX or maybe on all hardware but we never reached that condition on any other hardware.
So you understand my doubt.

I need to dig and check if the condition should be more accurately checked.

Anyway for now, can you change your PR following my comment below ?

lib/GLPI/Agent/Daemon.pm
phillscholl: done

g-bougard:
I think I was able to reproduce the problem using a test script.

With my test script, It finally appeared we may enter a loop if the forked process dies unexpectedly.

A simple fix would be to just call `last` on failed sysread, so replacing `$self->child_exit($pid);` by just `last;` is sufficient in that case. As I explained in #1041, glpi-agent still has all the logic to handle forked process. So we must continue to expect this part is doing the job.

I'll push a fix myself and you would be able to test with next nightly build.

g-bougard:
Closed by 60f011d

